PR #90, opened 2021-09-02
Automated testing

Brennan1994:


Brennan1994:
HEy Hey! we're working. Tests, and artifact upload. There's a lot of commits here because thats what I was using to trigger the build each time to see if I accomplished my goal. 

I did remove the OccupancyTypesTest Proj because It was empty and throwing a warning during the build that I suspected was causing it to register as a failure despite all tests passing. I no longer think that's the issue, but since we don't have any tests in it anyway, I feel ok with leaving it out until we actually populate it with something.

rnugent3 (approved):
Looks good!

Brennan1994:
> I wonder what this file was doing under occupancy type tests. Setting up to test getting the correct damage function?

There was no test written against it, so who knows what was planned. I'd guess that all the occupancy types were intended to be stored in a text file, and there was gonna be some test an importer for them. Or they were gonna be writeable to text file, and that was gonna be the reference, Maybe both.

Brennan1994:
> > I wonder what this file was doing under occupancy type tests. Setting up to test getting the correct damage function?
> 
> There was no test written against it, so who knows what was planned. I'd guess that all the occupancy types were intended to be stored in a text file, and there was gonna be some test an importer for them. Or they were gonna be writeable to text file, and that was gonna be the reference, Maybe both.

ehh, it might have been there was a failing test written against it that was deleted. Would have to check the history. Either way, I think we're staying true to our path by ignoring it for now till we need to address it again.

PR #160, opened 2021-10-19
Feature/deterministic distribution

rnugent3:


Statistics/Distributions/Deterministic.cs
```diff
@@ -0,0 +1,107 @@
...
+                return 1;
+            } else
+            {
+                return 0;
+            }
+        }
+
+        public bool Equals(IDistribution distribution)
+        {
+            throw new NotImplementedException();
```
HenryGeorgist: this needs to be implemented - check to see if x is the same.

Statistics/Distributions/Deterministic.cs
```diff
@@ -100,7 +98,9 @@ public String WriteToXML()
 
         XElement ISerializeToXML<IDistribution>.WriteToXML()
         {
-            throw new NotImplementedException();
+            XElement ordinateElem = new XElement(SerializationConstants.DETERMINISTIC);
+            ordinateElem.SetAttributeValue(SerializationConstants.CONSTANT_VALUE, Value);
```
HenryGeorgist, line 102: Your write and read should match.

Statistics/Distributions/Deterministic.cs
```diff
@@ -61,10 +62,7 @@ public double CDF(double x)
             }
         }
 
-        public bool Equals(IDistribution distribution)
-        {
-            throw new NotImplementedException();
-        }
+        public bool Equals(IDistribution distribution) => string.Compare(Print(), distribution.Print()) == 0 ? true : false;
```
HenryGeorgist, line 65: I think it should not use string compare, but rather just compare the x value.

PR #161, opened 2021-10-20
Feature/remove utils stats

Brennan1994:


.github/workflows/SimpleBuild.yml
```diff
@@ -33,32 +33,6 @@ jobs:
 
     - name: Add Source
       run: dotnet nuget add source --username HydrologicEngineeringCenter --password ${{ secrets.GITHUB_TOKEN }} --store-password-in-clear-text --name github "https://nuget.pkg.github.com/HydrologicEngineeringCenter/index.json"
-      
-    # Restore the application to populate the obj folder with RuntimeIdentifiers
```
rnugent3: Out of curiosity, why add this code in then delete?

Brennan1994: I only wanted to add the Add Source task to the script. It was a copy paste error that all the other stuff came with it.

PR #171, opened 2021-10-29
updating the description window and description text box to match con…

### LINKED ISSUE #166: Fix or create new Description Control
The “TextBoxDescription” control no longer works in FDA. I am not sure when it got broken. I vaguely remember you wrote this control many years ago. I could be wrong. It is written in VB and was taken from the consequences assist library. I have tried to figure out what is going wrong with it but have had no success. I first noticed that my description was not saving to the DB. While debugging I discovered that the binding to my “Description” property never gets hit. If you pop out the description window it just says “False” and if you try to type in it an exception gets thrown. Are you able to look into this? We use this control in most editors. You can look at the one in RatingCurveEditor.xaml. The view model for that class is the “CurveEditorVM” although the description property is down in “BaseEditorVM”.

HenryGeorgist:
…sequences assist fixes #166


codymccoy:
Cody, make sure to update the study description and the study notes in the study properties.

PR #214, opened 2021-12-15
Added code to update the inventory database table when an occtype edi…

codymccoy:
…tor change happens.


RyanM-RMA (approved):
Code reviewed, ready for test

PR #215, opened 2021-12-16
Bug fixes related to removing inventories and impact areas

codymccoy:
Bug fixes. The structure inventories were not being removed properly when deleting the child element. When investigating, other elements were also not deleting correctly. I added logic to delete the secondary table in the database and delete the geo package reference to the table in the geo package tables.

### DIFF
--- a/ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
+++ b/ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
@@ -218,6 +218,10 @@ namespace ViewModel.AggregatedStageDamage
                     int structID = CalculatedVM.SelectedStructures.GetElementID();
                     LastEditDate = DateTime.Now.ToString("G");
                     AggregatedStageDamageElement elem = new AggregatedStageDamageElement(Name, LastEditDate, Description, wseID, structID, ManualVM.GetStageDamageCurves(), true);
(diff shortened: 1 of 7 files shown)


ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
```diff
@@ -218,6 +218,10 @@ public virtual void SaveWhileEditing()
                     int structID = CalculatedVM.SelectedStructures.GetElementID();
                     LastEditDate = DateTime.Now.ToString("G");
                     AggregatedStageDamageElement elem = new AggregatedStageDamageElement(Name, LastEditDate, Description, wseID, structID, ManualVM.GetStageDamageCurves(), true);
+                    if (CurrentElement == null)
```
RyanM-RMA, line 221: Why do we only do this when the CurrentElement is null?

RyanM-RMA (commented):
Just one question I'd like to understand, and we should be good to go.

ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
RyanM-RMA: Spoke with Cody on a call to go over this.

RyanM-RMA (approved):

PR #235, opened 2021-12-27
Bug fixes for task 4

codymccoy:


Importer/AggregatedDamage/AggregateDamageFunction.cs
```diff
@@ -432,9 +427,12 @@ private AggregatedStageDamageElement CreateStageDamageElement(SingleDamageFuncti
...
             }
             //always use linear. This is the only option in Old Fda.
-            ICoordinatesFunction func = Functions.ICoordinatesFunctionsFactory.Factory(depthsList, damagesList, InterpolationEnum.Linear);
+            ICoordinatesFunction func = ICoordinatesFunctionsFactory.Factory(depthsList, damagesList, InterpolationEnum.Linear);
             IFdaFunction stageDamage = IFdaFunctionFactory.Factory( IParameterEnum.InteriorStageDamage, (IFunction)func);
-            AggregatedStageDamageElement elem = new AggregatedStageDamageElement(Name, CalculationDate, Description, stageDamage, CreationMethodEnum.Imported);
+            List<StageDamageCurve> curves = new List<StageDamageCurve>();
+            //todo: cody will fix in task 9
+            //StageDamageCurve curve = new StageDamageCurve()
+            AggregatedStageDamageElement elem = new AggregatedStageDamageElement(Name, CalculationDate, Description, -1,-1, curves, true);
```
RyanM-RMA, line 435: don't need the named variable, just return the new instance.

ViewModel/AggregatedStageDamage/AggregatedStageDamageOwnerElement.cs
```diff
@@ -77,13 +77,6 @@ public void AddNewStageDamageCurveSet(object arg1, EventArgs arg2)
 
         #endregion
         #region Functions
```
RyanM-RMA, line 79: Remove the region since it's empty

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
```diff
@@ -192,9 +192,28 @@ public List<StageDamageCurve> GetStageDamageCurves()
...
             bool areRowsUnique = AreManualRowsUniqueCombinations();
             bool curvesValid = AreManualCurvesValid();
-            return areRowsUnique && curvesValid;
+            return areRowsUnique && curvesValid && atLeast2Points;
+        }
+
+        private bool AreThereTwoPointsPerCurve()
+        {
+            bool atLeastTwoPoints = true;
+            foreach (ManualStageDamageRowItem r in Rows)
```
RyanM-RMA, line 205: I would do this differently, iterate over all the rows and keep track of the unique names of each one that fails the check.  then at the end when you show the message box, you list out the ones that were failing.  This will give the user a better understanding of what's failing and why.

ViewModel/Saving/PersistenceManagers/ImpactAreaPersistenceManager.cs
```diff
@@ -121,14 +121,24 @@ public void SaveNew(ChildElement element)
         {
             if (element.GetType() == typeof(ImpactAreaElement))
```
RyanM-RMA, line 122: This can be turned into pattern matching, instead of type == typeof

RyanM-RMA (changes requested):
Few things to change.

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
RyanM-RMA: Something to the effect of:
The following manually entered curves must have at least 2 points:
- Curve A
- Curve B
- ...

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
```diff
@@ -192,9 +192,28 @@ public List<StageDamageCurve> GetStageDamageCurves()
...
+            {
+                ICoordinatesFunction coordFunc = r.EditorVM.CreateFunctionFromTables();
+                if(coordFunc.Coordinates.Count<2)
+                {
+                    atLeastTwoPoints = false;
+                    String msg = "Mannually entered curves must have at least 2 points.";
```
RyanM-RMA, line 211: Manually is spelled wrong.

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
RyanM-RMA: Something like:
string.Join(Environment.NewLine + "\t* ", list)

where list is your list of curves.

codymccoy: I followed your advice and it now looks like this:
image

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
codymccoy: fixed

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
```diff
@@ -202,17 +202,22 @@ public bool ValidateForm()
...
-                    break;
                 }
             }
+
+            if(rowsThatFailed.Count>0)
+            {
+                String msg = "Manually entered curves must have at least 2 points." + Environment.NewLine + "Curves in error:" + Environment.NewLine + "\t* ";
+                MessageBox.Show(msg + string.Join(Environment.NewLine + "\t* ", rowsThatFailed), "Two Points Required", MessageBoxButton.OK, MessageBoxImage.Error);
+            }
             return atLeastTwoPoints;
```
RyanM-RMA, line 221: Instead of tracking and returning atLeastTwoPoints, I'd just return rowsThatFailed.Count == 0;  since that's your solid yes case.

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
RyanM-RMA: Could we go with ALT+0183 or \u2022 as the bullet point instead of *?  I find that it's cleaner looking than *.  If you don't think that's necessary that's fine.  It looks good as is but I think this adds polish.

ALT+0183 is a bit different than \u2022, \u2022 is an escape sequence so you just do "\t\u2022" instead of "\t*".  ALT+0183 needs to be typed by pressing and holding ALT, then typing the number, and letting go of ALT, and it'll add the character directly to the string instead of an escape sequence.  If you use \u2022, make a comment about how it's a bullet point character, because it helps clarify what the character is intended to be.

RyanM-RMA (changes requested):
Just a couple more things and I think you're good to go

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
codymccoy: Good idea. Fixed.

ViewModel/AggregatedStageDamage/ManualStageDamageVM.cs
codymccoy: I added the \u2022.
image

RyanM-RMA (approved):
Looks good to me

PR #253, opened 2022-01-27
Made the flow frequency write to xml and read from xml.

codymccoy:


RyanM-RMA (approved):
I need to review this and provide feedback to Cody

PR #263, opened 2022-02-07
Task1 and 2 fully implement

codymccoy:
This code is integrating the new model and removing the old model, functions. 
This is mostly code clean up and changing old terms for new terms. This was my cheat sheet:
image

Some classes/areas of code i didn't bother "cleaning" because they will be addressed in another task down the road.


RyanM-RMA (changes requested):
Found some minor issues, but the dead/zombie code comments can be resolved if you think the code is still useful as we discussed.

RyanM-RMA (approved):

PR #271, opened 2022-02-14
Task4 review updates

codymccoy:
These are some task review fixes requested by Richard. This is for tasks 4 and 5.

### DIFF
--- a/View/Inventory/StructureMissingElevationEditor.xaml
+++ b/View/Inventory/StructureMissingElevationEditor.xaml
@@ -13,6 +13,7 @@
         <con:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter"/>
         <con:BoolToInverseVisibilityConverter x:Key="BoolToInverseVisConverter"/>
         <con:BoolToMissingConverter x:Key="BoolToMissing"/>
+        <con:BoolToUnidentifiableConverter x:Key="BoolToUnidentifiable"/>
     </UserControl.Resources>
     <Grid>
         <Grid.Resources>
@@ -54,7 +55,7 @@
                 
                 <DataGridTextColumn Binding="{Binding IsMissingFirstFloorElevation, Converter={StaticResource BoolToMissing}}" Header="First Floor Elevation"  Width="*" Visibility="{Binding DataContext.UsingFirstFloorElevation, Converter={StaticResource BoolToVisibilityConverter}, Source={x:Reference dummyElement}}"/>
                 <DataGridTextColumn Binding="{Binding IsMissingStructureValue, Converter={StaticResource BoolToMissing}}" Header="Structure Value"  Width="*"/>
-                <DataGridTextColumn Binding="{Binding IsMissingTerrainElevation, Converter={StaticResource BoolToMissing}}" Header="Terrain Elevation"  Width="*" Visibility="{Binding DataContext.ElevsFromTerrainFile, Converter={StaticResource BoolToVisibilityConverter}, Source={x:Reference dummyElement}}" />
+                <DataGridTextColumn Binding="{Binding IsMissingTerrainElevation, Converter={StaticResource BoolToUnidentifiable}}" Header="Terrain Elevation"  Width="*" Visibility="{Binding DataContext.ElevsFromTerrainFile, Converter={StaticResource BoolToVisibilityConverter}, Source={x:Reference dummyElement}}" />
 
             </DataGrid.Columns>
         </DataGrid>
--- a/View/Utilities/BoolToMissingConverter.cs
+++ b/View/Utilities/BoolToMissingConverter.cs
@@ -8,7 +8,7 @@ namespace View.Utilities
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            if ((bool)value == true)
+            if (System.Convert.ToBoolean(value) == true)
             {
                 return "Missing";
             }
--- a/View/Utilities/BoolToUnidentifiableConverter.cs
+++ b/View/Utilities/BoolToUnidentifiableConverter.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace View.Utilities
+{
+    public class BoolToUnidentifiableConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if ((bool)value == true)
+            {
+                return "Unidentifiable";
+            }
+            else
+            {
+                return null;
+            }
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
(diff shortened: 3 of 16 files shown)


ViewModel/Utilities/ImportFromFDA1Helper.cs
```diff
@@ -57,108 +57,110 @@ private static RatingCurveElement CreateRatingElement(RatingFunction rat)
...
-            double[] stages = rat.GetStage();
-            double[] flows = rat.GetDischarge();
-
-            if (rat.ErrorTypesId == ErrorType.LOGNORMAL)
+            List<IDistribution> ys = new List<IDistribution>();
+            for (int i = 0; i < rat.NumberOfPoints; i++)
             {
-                List<IDistribution> ys = new List<IDistribution>();
-                for (int i = 0; i < rat.NumberOfPoints; i++)
+                if (rat.UsesGlobalError)
```
RyanM-RMA, line 91: Not a huge fan of the name, I'd say UsesGlobalStd or something more descriptive, but I don't know its full usage outside of these cases.

ViewModel/Utilities/ImportFromFDA1Helper.cs
```diff
@@ -57,108 +57,110 @@ private static RatingCurveElement CreateRatingElement(RatingFunction rat)
...
+                {
+                    ys.Add(new LogNormal(rat.GetDischarge()[i], rat.GlobalStdDevLog));
+                }
+                else
                 {
-                    ys.Add(new LogNormal(flows[i], rat.GlobalStdDev));
+                    ys.Add(new LogNormal(rat.GetDischarge()[i], rat.IndividualLogStDevs[i]));
```
RyanM-RMA, line 97: Any way we can make the property names consistent between GlobalStd's and IndividualStd's?  Globals are consistent with Globals, and Individuals are consistent with Individuals, but they aren't consistent with each other.  It makes it more confusing and inconsistent when you have to use one or the other based on an arbitrary bool property that doesn't refer to Std in the name.

ViewModel/Utilities/ImportFromFDA1Helper.cs
RyanM-RMA: Doing this logic on the bool property feels very arbitrary from an end user of the API standpoint.  Is there a way we can do this logic in the API at all?  Probably not, I see that we have two different types for the output.  Just a thought that's probably not actionable.

RyanM-RMA (changes requested):
There are some changes requested.  Some API considerations in the RatingFunction code (I think that's in the new Function API right?)

View/Inventory/StructureMissingElevationEditor.xaml
```diff
@@ -54,7 +55,7 @@
                 
                 <DataGridTextColumn Binding="{Binding IsMissingFirstFloorElevation, Converter={StaticResource BoolToMissing}}" Header="First Floor Elevation"  Width="*" Visibility="{Binding DataContext.UsingFirstFloorElevation, Converter={StaticResource BoolToVisibilityConverter}, Source={x:Reference dummyElement}}"/>
                 <DataGridTextColumn Binding="{Binding IsMissingStructureValue, Converter={StaticResource BoolToMissing}}" Header="Structure Value"  Width="*"/>
-                <DataGridTextColumn Binding="{Binding IsMissingTerrainElevation, Converter={StaticResource BoolToMissing}}" Header="Terrain Elevation"  Width="*" Visibility="{Binding DataContext.ElevsFromTerrainFile, Converter={StaticResource BoolToVisibilityConverter}, Source={x:Reference dummyElement}}" />
+                <DataGridTextColumn Binding="{Binding IsMissingTerrainElevation, Converter={StaticResource BoolToUnidentifiable}}" Header="Terrain Elevation"  Width="*" Visibility="{Binding DataContext.ElevsFromTerrainFile, Converter={StaticResource BoolToVisibilityConverter}, Source={x:Reference dummyElement}}" />
```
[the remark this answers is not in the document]
codymccoy: yes it is still being used by the other columns.

ViewModel/Utilities/ImportFromFDA1Helper.cs
```diff
@@ -57,108 +57,110 @@ private static RatingCurveElement CreateRatingElement(RatingFunction rat)
 
         private static UncertainPairedData CreateRatingPairedData(RatingFunction rat)
         {
-            if (rat.UsesGlobalError)
+            List<IDistribution> ys = new List<IDistribution>();
+
+            if (rat.ErrorTypesId == ErrorType.LOGNORMAL)
```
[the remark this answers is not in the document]
codymccoy: good idea. I made it a switch.

ViewModel/Utilities/ImportFromFDA1Helper.cs
codymccoy: This is coming from Bob's importer code. I don't really want to dive into that code and change. Things. Lets just leave this alone for now.

ViewModel/Utilities/ImportFromFDA1Helper.cs
codymccoy: Yes, this is all kind of weird but comes from Bob's importer code. He is distinguishing if the uncertainty value applies to the whole curve or if there is uncertainty around each coordinate in the curve.

View/Utilities/BoolToMissingConverter.cs
```diff
@@ -8,7 +8,7 @@ public class BoolToMissingConverter : IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            if ((bool)value == true)
+            if (System.Convert.ToBoolean(value) == true)
```
RyanM-RMA, line 11: Namespacing, you don't need System.

View/Utilities/BoolToUnidentifiableConverter.cs
```diff
@@ -8,7 +8,7 @@ public class BoolToUnidentifiableConverter : IValueConverter
     {
         public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
-            if ((bool)value == true)
+            if (System.Convert.ToBoolean(value) == true)
```
RyanM-RMA, line 11: Namespacing, you don't need System.

RyanM-RMA (approved):
I'm approving it, but you don't need the System namespace in System.Convert.ToBoolean, you just want Convert.ToBoolean.

View/Utilities/BoolToMissingConverter.cs
codymccoy: You need "System" because the method name is Convert and it gets confused.

View/Utilities/BoolToUnidentifiableConverter.cs
codymccoy: You need "System" because the method name is Convert and it gets confused.

PR #309, opened 2022-03-24
occtype editor bug fixes.

codymccoy:


RyanM-RMA (changes requested):
Some zombie code, but it looks good

RyanM-RMA (approved):

PR #312, opened 2022-04-01
Moved the "removing" logic down to the child element class and fixed …

codymccoy:
…some bugs in the hydraulics.

### DIFF
--- a/ViewModel/ImpactArea/ImpactAreaElement.cs
+++ b/ViewModel/ImpactArea/ImpactAreaElement.cs
@@ -69,35 +69,36 @@ namespace HEC.FDA.ViewModel.ImpactArea
 
         public void RemoveElement(object sender, EventArgs e)
         {
-            MessageBoxResult messageBoxResult = MessageBox.Show("Are you sure you want to delete '" + Name + "'?", "Delete " + Name + "?", MessageBoxButton.YesNo, MessageBoxImage.Question);
-            if (messageBoxResult == MessageBoxResult.Yes)
+            List<IASElementSet> iasElems = StudyCache.GetChildElementsOfType<IASElementSet>();
+            if (iasElems.Count > 0)
             {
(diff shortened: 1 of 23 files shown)


ViewModel/ImpactArea/ImpactAreaElement.cs
```diff
@@ -69,30 +69,34 @@ public ImpactAreaElement(string userdefinedname,string description, ObservableCo
 
         public void RemoveElement(object sender, EventArgs e)
         {
-            List<IASElementSet> iasElems = StudyCache.GetChildElementsOfType<IASElementSet>();
-            if (iasElems.Count > 0)
+            MessageBoxResult messageBoxResult = MessageBox.Show("Are you sure you want to delete '" + Name + "'?", "Delete " + Name + "?", MessageBoxButton.YesNo, MessageBoxImage.Question);
```
RyanM-RMA, line 72: As we discussed, let's make this one dialog per scenario instead of two dialogs if you have multiple IASElementSets

ViewModel/WaterSurfaceElevation/WaterSurfaceElevationElement.cs
```diff
@@ -145,7 +150,7 @@ public override void RemoveElementFromMapWindow(object arg1, EventArgs arg2)
             }
         }
 
-        public void removedcallback(OpenGLMapping.FeatureNodeHeader node, bool includeSelected)
+        public void removedcallback(FeatureNodeHeader node, bool includeSelected)
```
RyanM-RMA, line 153: Rename to RemovedCallback

RyanM-RMA (changes requested):
Let's change the delete confirmation for IASElement so it only asks once, and there's a function rename requested.

RyanM-RMA (changes requested):
so close!  one more change please.

RyanM-RMA (approved):

PR #323, opened 2022-04-13
Bug/table with plot validation

Brennan1994:


HenryGeorgist (approved):
looks good. id question if the pasting logic is robust to deal with issues where properties are out of order (i.e. what if x is not the last property).

Brennan1994:
> looks good. id question if the pasting logic is robust to deal with issues where properties are out of order (i.e. what if x is not the last property).

I'm nervous about that too, but I'm unsure how to address it right now. I think it has something to do with how we've set up the indexing of columns on the display as column attribute. even though we display X at index 0, we do not put X at index 0. This fix got us away from referencing sequential row directly, but still leaves some room for improvement.  I'm open to suggestions if you have ideas.

PR #352, opened 2022-05-19
Fixing bug where import tab/window gets removed when importing a study.

codymccoy:


ViewModel/Tabs/TabController.cs
```diff
@@ -263,11 +265,20 @@ private void PopTabIntoWindow(object sender, EventArgs e)
         public void CloseTabsAndWindowsOpeningNewStudy()
         {
             _Windows.Clear();
-            _Tabs.Clear();
+
+            //remove all the tabs except for the importing tab
+            List<IDynamicTab> dynamicTabsToRemove = _Tabs.Where(tab => !tab.UniqueName.Equals(StudyElement.IMPORT_STUDY_UNIQUE_STRING)).ToList();
```
RyanM-RMA, line 270: I recommend using _Tabs.RemoveAll(tab => ...);

It takes in your predicate, and removes the element if the predicate returns true.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeall?view=net-6.0

RyanM-RMA (changes requested):
Small request

RyanM-RMA (approved):

PR #354, opened 2022-05-20
Bug/copy paste

Brennan1994:
@HenryGeorgist and I worked through this this morning. Copy paste now behaves as it should when pasting a column, and not a whole grid.

### DIFF
--- a/View/TableWithPlot/FdaDataGrid.cs
+++ b/View/TableWithPlot/FdaDataGrid.cs
@@ -267,7 +267,6 @@ namespace HEC.FDA.View.TableWithPlot
                     DataGridCellInfo cellinfo = SelectedCells[0];
                     RowIndex = Items.IndexOf(cellinfo.Item);
                     ColumnIndex = cellinfo.Column.DisplayIndex;
-                    //so this is tricky, because when we sift through the properties on the row objects, it's always going to be X last, but in our clipboards, it'll always be the unique properties first, then X. so I need to skip the first 
                     for (Int32 i = 0; i <= clipboardData.Count() - 1; i++)
                     {
                         if ((RowIndex + i) > Items.Count - 1)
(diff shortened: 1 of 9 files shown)


View/TableWithPlot/FdaDataGrid.cs
```diff
@@ -293,13 +292,15 @@ public void PasteClipboard()
...
                             {
-                                y.SetValue(Items[RowIndex + i], Convert.ChangeType(clipboardData[i][0], y.PropertyType));
+                                var newVal = Convert.ChangeType(clipboardData[i][j], y.PropertyType);
+                                ((SequentialRow)Items[RowIndex + i]).UpdateRow(ColumnIndex + j, (double)newVal);
                             }
-                            else
+                            catch (Exception ex)
                             {
-                                y.SetValue(Items[RowIndex + i], Convert.ChangeType(clipboardData[i][j + 1], y.PropertyType));
+                                //ignore
```
rnugent3, line 303: Are the potential exceptions not worth handling?

ViewModel/TableWithPlot/Rows/GraphicalRow.cs
```diff
@@ -10,7 +10,37 @@ public class GraphicalRow : SequentialRow
...
+            }
+            set
+            {
+                Y = new Deterministic(value);
+                Confidence05 = 0;
+                Confidence95 = 0;
+                NotifyPropertyChanged();
+            }
+        }
         [DisplayAsColumn("5% Confidence")]
```
rnugent3, line 44: I know this is not new code, but would just like to mention that the field is used to a 95% confidence interval. If we're trying to maintain as smooth of a transition as practicable, then I think a 95% interval would be better.

View/TableWithPlot/FdaDataGrid.cs
```diff
@@ -448,10 +449,10 @@ private void myDataGrid_AutoGeneratingColumn(object sender, DataGridAutoGenerati
                             if (dna != null)
                             {
                                 Console.WriteLine(dna.DisplayName);
-                                if(dna.DisplayIndex >= 0)
-                                {
-                                    dgtc.DisplayIndex = dna.DisplayIndex;
-                                }
+                                //if(dna.DisplayIndex >= 0)
```
rnugent3, line 452: if we'd like to keep commented out code, it would be helpful to include some comments on why we've left the commented out code in the class.

rnugent3 (approved):
Great work, guys. Thank you.

View/TableWithPlot/FdaDataGrid.cs
Brennan1994: Recycled code in here. Looks like the original author didn't think so. I'd be ok accepting it as is for now until we find an issue with it.

ViewModel/TableWithPlot/Rows/GraphicalRow.cs
Brennan1994: Easily done. Right now this class isn't linked to a view model. I'll make sure that lines up when I move that in.

View/TableWithPlot/FdaDataGrid.cs
Brennan1994: agreed. gone.

PR #406, opened 2022-06-22
Feature/stage frequency

Brennan1994:
Incorporating Stage frequency into the graphical stuff

### DIFF
--- a/ViewModel/FrequencyRelationships/GraphicalVM.cs
+++ b/ViewModel/FrequencyRelationships/GraphicalVM.cs
@@ -13,9 +13,12 @@ namespace HEC.FDA.ViewModel.FrequencyRelationships
     public class GraphicalVM : ComputeComponentVM
     {
         private int _equivalentRecordLength = 5;
+        private bool _useStage;
+        private bool _useFlow = true;
         private NamedAction _confidenceLimits;
-        public NamedAction ConfidenceLimits { get { return _confidenceLimits; } set { _confidenceLimits = value; NotifyPropertyChanged(); } }
 
+
+        public NamedAction ConfidenceLimits { get { return _confidenceLimits; } set { _confidenceLimits = value; NotifyPropertyChanged(); } }
         public GraphicalUncertainPairedData MyGraphical
         {
             get{return new GraphicalUncertainPairedData(((GraphicalDataProvider)SelectedItem).Xs, ((GraphicalDataProvider)SelectedItem).Ys, EquivalentRecordLength,new CurveMetaData(), usingStagesNotFlows: true);}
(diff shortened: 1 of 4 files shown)


Brennan1994:
@codymccoy hoping you can take a look at the incorporation into the compute and verify that looks reasonable to you (the last commit)

ViewModel/FrequencyRelationships/GraphicalVM.cs
```diff
@@ -13,9 +13,12 @@ namespace HEC.FDA.ViewModel.FrequencyRelationships
     public class GraphicalVM : ComputeComponentVM
     {
         private int _equivalentRecordLength = 5;
+        private bool _useStage;
+        private bool _useFlow = true;
         private NamedAction _confidenceLimits;
-        public NamedAction ConfidenceLimits { get { return _confidenceLimits; } set { _confidenceLimits = value; NotifyPropertyChanged(); } }
 
+
+        public NamedAction ConfidenceLimits { get { return _confidenceLimits; } set { _confidenceLimits = value; NotifyPropertyChanged(); } }
```
codymccoy, line 21: It is best to put getters and setters on different lines. This becomes helpful when needing to put breakpoints on them.

ViewModel/TableWithPlot/ComputeComponentVM.cs
```diff
@@ -55,6 +55,7 @@ public string XLabel
             {
                 _xlabel = value;
                 NotifyPropertyChanged();
+                NotifyPropertyChanged(nameof(Units));
```
codymccoy, line 58: Why add another notify prop changed here?

ViewModel/TableWithPlot/ComputeComponentVM.cs
```diff
@@ -67,6 +68,7 @@ public string YLabel
             {
                 _ylabel = value;
                 NotifyPropertyChanged();
+                NotifyPropertyChanged(nameof(Units));
```
codymccoy, line 71: and here?

codymccoy (commented):
A couple little things

ViewModel/TableWithPlot/ComputeComponentVM.cs
Brennan1994: The label on the UI is bound to the property "Units" which takes XLabel and YLabel and formats them into a nice string to display. I update Y label when I switch between stage and discharge, so i need units to change when y label changes

ViewModel/TableWithPlot/ComputeComponentVM.cs
Brennan1994: same as above

PR #407, opened 2022-06-23
Fda bug fixes

codymccoy:


RyanM-RMA (changes requested):
If they're not zombies that's ok, but get rid of the ones that are.

RyanM-RMA (approved):

PR #413, opened 2022-06-27
Fda testing

codymccoy:


View/ImpactAreaScenario/Results/PerformanceAEP.xaml
```diff
@@ -11,15 +11,30 @@
...
+            <Grid.RowDefinitions>
+                <RowDefinition Height="Auto"/>
+                <RowDefinition Height="Auto"/>
+            </Grid.RowDefinitions>
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="Auto"/>
+                <ColumnDefinition Width="*"/>
+            </Grid.ColumnDefinitions>
+            <Label Grid.Row="0" Grid.Column="0" Content="Mean:" HorizontalAlignment="Right" />
+            <TextBlock Grid.Row="0" Grid.Column="1" Text="{Binding Mean, StringFormat={x:Static utilVM:StringConstants.DEFAULT_UNIT_FORMAT}}" VerticalAlignment="Center"/>
```
RyanM-RMA, line 32: Are these intended to be editable?

ViewModel/ImpactAreaScenario/Editor/SpecificIASEditorVM.cs
```diff
@@ -516,7 +516,7 @@ public void MyMessageHandler(object sender, MessageEventArgs e)
             if (e.Message is FrequencyDamageMessage damageMessage)
             {
                 //todo: not sure that this is correct. Maybe we want the "total" one, but in the current case the "total" has no values?
-                if(e.Message.Message.Equals("Damage-frequency function for damage and asset categoriesdefaultandunassigned"))
+                if(e.Message.Message.Equals("FrequencyDamage"))
```
RyanM-RMA, line 519: you can use damageMessage.Message instead of e.Message.Message.

RyanM-RMA: Is there a constant for FrequencyDamage somewhere that you can reference?

ViewModel/ImpactAreaScenario/Results/ThresholdComboItem.cs
```diff
@@ -1,4 +1,5 @@
-﻿using metrics;
+﻿using HEC.FDA.ViewModel.Utilities;
```
RyanM-RMA, line 1: Unused import

RyanM-RMA (changes requested):

View/ImpactAreaScenario/Results/PerformanceAEP.xaml
codymccoy: No. These are text blocks and not text boxes. They can't be edited.

ViewModel/ImpactAreaScenario/Editor/SpecificIASEditorVM.cs
codymccoy: I don't think so. This is a message that comes out of the compute, so this string is defined somewhere in HEC model. The message has changed a couple of times. I think HEC has enough on their plate right now, but ideally this would be handled better.

RyanM-RMA (approved):

ViewModel/ImpactAreaScenario/Editor/SpecificIASEditorVM.cs
HenryGeorgist: i think if you cast, and then use the FrequencyDamageMessage you have access to a const rather than parsing the message string.

HenryGeorgist: https://github.com/HydrologicEngineeringCenter/fda-model/blob/597a1a24bbc7a778a6e0b186d477bcd8a4b0528f/fda-model/compute/FrequencyDamageMessage.cs#L21

then dig into the paired data object metadata

PR #414, opened 2022-06-27
Feature/graphical from old fda

Brennan1994:


Brennan1994:
With this Graphical should be able to be imported from FDA 1.X

rnugent3 (approved):
some small things that could be recorded and addressed later

PR #417, opened 2022-06-28
Bug/stage flow graphical

Brennan1994:


ViewModel/FrequencyRelationships/GraphicalVM.cs
```diff
@@ -150,7 +152,8 @@ private void ConfidenceLimitsAction(object arg1, EventArgs arg2)
         }
         public GraphicalUncertainPairedData ToGraphicalUncertainPairedData()
         {
-           return new GraphicalUncertainPairedData(((GraphicalDataProvider)SelectedItem).Xs, ((GraphicalDataProvider)SelectedItem).Ys , EquivalentRecordLength, "Exceedence Probability", "Flow", "Graphical Flow Frequency");
+            CurveMetaData meta = new CurveMetaData(Utilities.StringConstants.EXCEEDANCE_PROBABILITY,Utilities.StringConstants.DISCHARGE,Utilities.StringConstants.GRAPHICAL_FLOW_FREQUENCY,"None",CurveTypesEnum.MonotonicallyIncreasing);
```
codymccoy, line 155: It is best to not use namespaces. You can remove all of these "Utilities".

Brennan1994: Other references become ambiguous when I use a using statement because I'm using a namespace in Will's framework called Utilities. Until that's fixed, which I think we're stuck here.

PR #419, opened 2022-06-28
Fixing bug. User can remove element while it is being edited.

codymccoy:


ViewModel/ImpactAreaScenario/Editor/SpecificIASEditorVM.cs
```diff
@@ -335,7 +335,16 @@ private void UpdateRatingRequired()
         {
             if (SelectedFrequencyElement != null && SelectedFrequencyElement.ChildElement is AnalyticalFrequencyElement elem)
             {
-                RatingRequired = !elem.IsAnalytical;
+                RatingRequired = false;
+                if(elem.IsAnalytical)
```
RyanM-RMA, line 339: I recommend RatingRequired = elem.IsAnalytical || elem.MyGraphicalVM.UseFlow or collapsing the if block.

RyanM-RMA (changes requested):
Minor change with your if statement

ViewModel/ImpactAreaScenario/Editor/SpecificIASEditorVM.cs
codymccoy: This is hard to do, it is as simple as your suggestion because MyGraphicalVM always has values. Bottom line is i think it is simpler to read the way that it is.

PR #437, opened 2022-07-11
Bug fixes

codymccoy:


Brennan1994:
from @codymccoy  "You mentioned switching the copy paste data grids over to FdaDataGrid. I switched all of them over except for the one in the analytical frequency editor. This one didn’t work because I am overriding some events that don’t exist on FdaDataGrid. We can look at that one down the road if we want to"

Merging pull request.

PR #438, opened 2022-07-14
Stage damage update

codymccoy:
This PR completes the stage damage update described in this document:
https://www.hec.usace.army.mil/confluence/display/FDA/2022-07-11+Maintenance+Tasks+Discussion

### DIFF
--- a/View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
+++ b/View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
@@ -54,7 +54,7 @@
             </Grid>
 
                 <ListView Grid.Row="0" Grid.Column="2"  Margin="5" ItemsSource="{Binding ImpactAreaFrequencyRows}" VerticalAlignment="Stretch"  HorizontalAlignment="Stretch" 
-                           SelectionMode="Single" ScrollViewer.VerticalScrollBarVisibility="Auto" SizeChanged="ListView_SizeChanged" >
+                           SelectionMode="Single" ScrollViewer.VerticalScrollBarVisibility="Auto" SizeChanged="ImpactAreaFrequencyListView_SizeChanged" >
                     <ListView.Resources>
                         <Style TargetType="ListViewItem">
                             <Setter Property="HorizontalContentAlignment" Value="Stretch" />
@@ -100,7 +100,7 @@
 
         <GroupBox Grid.Row="2" Grid.Column="0"  Header="Generated Curves" Margin="10,5,5,5" HorizontalAlignment="Stretch">
             <ListView  Margin="5" ItemsSource="{Binding Rows}" SelectedItem="{Binding SelectedRow}" SelectionMode="Single" 
-                       ScrollViewer.VerticalScrollBarVisibility="Auto" SizeChanged="ListView_SizeChanged_1">
+                       ScrollViewer.VerticalScrollBarVisibility="Auto" SizeChanged="ComputedCurves_SizeChanged">
                 <ListView.View>
                     <GridView>
                         <GridViewColumn Header="ID" Width="30" DisplayMemberBinding="{Binding ID}" />
--- a/View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
+++ b/View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
@@ -12,8 +12,8 @@
         <utils:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
         <local:StageDischargeVisibilityConverter x:Key="StageDischargeVisibilityConverter" />
         <local:InverseStageDischargeVisibilityConverter x:Key="InverseStageDischargeVisibilityConverter" />
-
     </UserControl.Resources>
+    
     <Grid>
         <Grid.RowDefinitions>
             <RowDefinition Height="*" MinHeight="30"/>
@@ -22,7 +22,6 @@
             <RowDefinition Height="5" />
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
-
 
         <GroupBox Header="Configure and Compute" Grid.Row="0" Margin="5">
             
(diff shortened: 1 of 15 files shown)


View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
```diff
@@ -4,45 +4,103 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
              xmlns:utils="clr-namespace:HEC.FDA.View.Utilities"
+             xmlns:local="clr-namespace:HEC.FDA.View.AggregatedStageDamage"
              mc:Ignorable="d" 
              xmlns:twp="clr-namespace:HEC.FDA.View.TableWithPlot"
              d:DesignHeight="450" d:DesignWidth="800">
     <UserControl.Resources>
         <utils:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
+        <local:StageDischargeVisibilityConverter x:Key="StageDischargeVisibilityConverter" />
```
Brennan1994, line 13: Would a Microsoft default converter "BooleanToVisibilityConverter" (System.Windows.Controls.BooleanToVisibilityConverter) satisfy your need here without requiring special classes?

View/AggregatedStageDamage/CalculatedStageDamageControl.xaml.cs
```diff
@@ -19,7 +19,57 @@ private void calculate_btn_Click(object sender, RoutedEventArgs e)
         {
             if (DataContext is CalculatedStageDamageVM vm)
             {
-                vm.CalculateCurves();
+                vm.ComputeCurves();
+            }
+        }
+
+        /// <summary>
+        /// This is to stretch the listview columns to fill the space as the user stretches the window.
```
Brennan1994, line 27: Is there a way to allow the columns to expand with the window by using wildcards in Xaml?

View/AggregatedStageDamage/CalculatedStageDamageControl.xaml.cs
```diff
@@ -19,7 +19,57 @@ private void calculate_btn_Click(object sender, RoutedEventArgs e)
...
+        /// <param name="e"></param>
+        private void ListView_SizeChanged(object sender, SizeChangedEventArgs e)
+        {
+            ListView listView = sender as ListView;
+            GridView gView = listView.View as GridView;
+            
+            // take into account vertical scrollbar
+            var workingWidth = listView.ActualWidth - SystemParameters.VerticalScrollBarWidth - 5; 
+            var col1 = 0.33;
+            var col2 = 0.33;
```
Brennan1994, line 39: if they're all .33, can we just use one variable?

View/AggregatedStageDamage/CalculatedStageDamageControl.xaml.cs
```diff
@@ -19,7 +19,57 @@ private void calculate_btn_Click(object sender, RoutedEventArgs e)
...
+                gView.Columns[2].Width = workingWidth * col3;
+            }
+        }
+
+        /// <summary>
+        /// This is to stretch the listview columns to fill the space as the user stretches the window.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void ListView_SizeChanged_1(object sender, SizeChangedEventArgs e)
```
Brennan1994, line 55: Is this the same code duplicated with a _1?

View/AggregatedStageDamage/InverseStageDischargeVisibilityConverter.cs
```diff
@@ -0,0 +1,34 @@
+﻿using HEC.FDA.ViewModel.AggregatedStageDamage;
+using System;
+using System.Globalization;
+using System.Windows;
+using System.Windows.Data;
+
+namespace HEC.FDA.View.AggregatedStageDamage
```
Brennan1994, line 7: See commend about BooleanToVisibilityConverter

View/AggregatedStageDamage/StageDischargeVisibilityConverter.cs
```diff
@@ -0,0 +1,34 @@
+﻿using HEC.FDA.ViewModel.AggregatedStageDamage;
```
Brennan1994, line 1: Same concern here. Can we do this with Boolean to visibility converter.

ViewModel/AggregatedStageDamage/AggregatedStageDamageElement.cs
```diff
@@ -17,15 +17,17 @@ public class AggregatedStageDamageElement : ChildElement
...
         public int SelectedStructures { get; set; }
-
         public List<StageDamageCurve> Curves { get; }
         public bool IsManual { get; }
+        public List<ImpactAreaFrequencyFunctionRowItem> ImpactAreaFrequencyRows { get; }
 
         #endregion
         #region Constructors
 
-        public AggregatedStageDamageElement(String name, string lastEditDate, string description,int selectedWSE, int selectedStructs, List<StageDamageCurve> curves, bool isManual, int id) : base(id)
```
Brennan1994, line 27: Rather than storing all the properties in the Stage damage element, can we just store the VM, like we did with Graphical? in the AnalyticalFlowFrequencyElement ?

ViewModel/AggregatedStageDamage/FrequencyElementWrapper.cs
```diff
@@ -0,0 +1,39 @@
+﻿using HEC.FDA.ViewModel.FrequencyRelationships;
+
+namespace HEC.FDA.ViewModel.AggregatedStageDamage
+{
+    /// <summary>
+    /// This class exists for the sole purpose of having a blank row in the combobox
```
Brennan1994, line 6: I'd rather not just hand jam something weird in. Is there a better way to do this?

ViewModel/AggregatedStageDamage/StageDischargeElementWrapper.cs
```diff
@@ -0,0 +1,36 @@
+﻿using HEC.FDA.ViewModel.StageTransforms;
+
+namespace HEC.FDA.ViewModel.AggregatedStageDamage
+{
+    /// <summary>
+    /// This class exists for the sole purpose of having a blank row in the combobox
+    /// </summary>
+    public class StageDischargeElementWrapper
```
Brennan1994, line 8: I'd rather not make these wrapper classes

Brennan1994 (changes requested):
My primary concerns are the wrapper classes, which are confusing. There has to be a cleaner way to handle that. I know this is probably the quickest way, but it leaves a lot of confusing code in it's wake that is going to really throw off anyone else who ever touches this code base. 

It'd be nice to use a default Bool to Viz converter too if that's possible. I think there's an inverse too. That'd save us another couple classes. 

Finally I trust your element works, but there's a lot of properties on it. For the Flow Frequency one, it seemed a little cleaner for graphical where the element just stored a whole graphical VM. Could we use that approach here? I have not fully explored that area of the code, and understand if there's nuance here I'm not seeing.

View/AggregatedStageDamage/CalculatedStageDamageControl.xaml.cs
codymccoy: done

View/AggregatedStageDamage/CalculatedStageDamageControl.xaml.cs
codymccoy: yes, these methods are auto generated events from the .xaml class. I mothed the central logic into its own method and call it from the two events. I changed the names to be more clear about which goes to which list view.

PR #439, opened 2022-07-14
Frequency radio bug

codymccoy:


View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
```diff
@@ -4,45 +4,103 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
              xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
              xmlns:utils="clr-namespace:HEC.FDA.View.Utilities"
+             xmlns:local="clr-namespace:HEC.FDA.View.AggregatedStageDamage"
              mc:Ignorable="d" 
              xmlns:twp="clr-namespace:HEC.FDA.View.TableWithPlot"
              d:DesignHeight="450" d:DesignWidth="800">
     <UserControl.Resources>
         <utils:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter" />
+        <local:StageDischargeVisibilityConverter x:Key="StageDischargeVisibilityConverter" />
```
Brennan1994, line 13: The StagedischargeVisibilityConverter isn't actually wired to Stage and discharge, you're just checking Bools. Could it be made more generic, or potentially just replaced with a bool to vis?

View/FrequencyRelationships/GraphicalControl.xaml.cs
```diff
@@ -9,5 +9,10 @@ public GraphicalControl()
             InitializeComponent();
             
         }
+
+        private void UserControl_DataContextChanged(object sender, System.Windows.DependencyPropertyChangedEventArgs e)
```
Brennan1994, line 13: What's this guy doing here? Is this necessary to fix the bug with the radio buttons?

View/FrequencyRelationships/GraphicalTableWithPlotControl.xaml
```diff
@@ -12,7 +14,33 @@
             <ColumnDefinition/>
             <ColumnDefinition/>
         </Grid.ColumnDefinitions>
+        
+        <!--<Grid Grid.Column="0">
```
Brennan1994, line 18: Delete commented out code.

View/FrequencyRelationships/GraphicalTableWithPlotControl.xaml
```diff
@@ -12,7 +14,33 @@
...
+                <RadioButton GroupName="FlowOrStage" Content="Flow" IsChecked="{Binding Path=ComputeComponentVM.UseFlow, Mode=TwoWay}" Margin="5"/>
+                <RadioButton GroupName="FlowOrStage" Content="Stage" IsChecked="{Binding Path=ComputeComponentVM.UseStage, Mode=OneWay}" Margin="5"/>
+            </StackPanel>
+            <twp:FdaDataGridControl DataContext="{Binding Path =ComputeComponentVM.SelectedItem }" Grid.Row="3"/>
+        </Grid>-->
+        
+        
         <frequencyrelationships:GraphicalControl DataContext="{Binding Path=ComputeComponentVM}" Grid.Column="0"/>
-        <oxy:PlotView Model="{Binding Path= PlotModel}" Grid.Column="1"/>
+        <oxy:PlotView Model="{Binding Path= ComputeComponentVM.PlotModel}" Grid.Column="1"/>
```
Brennan1994, line 44: This seems like it shouldn't work. Plot model isn't in the ComputeComponentVM, it's in the GraphicalTableWithPlotVM. I suspect this broke the plot.

Brennan1994 (changes requested):

View/AggregatedStageDamage/CalculatedStageDamageControl.xaml
codymccoy: This could be refactored to bind the visibility to a bool in the VM but then i think you would have to have all the same logic that is in my converter in the VM class. Somewhere you have to drill into the frequency object and check that bool. I chose to keep it all in the View rather than in the VM. I don't see a way to easily make this cleaner. To do it in the vm, I think you would have to put a method call in the setter of the selected frequency property to update the bool visibility property. I try to avoid calling things from the setters if i can. I view both approaches as equal and since this one is already written, i vote to keep it the way that it is.

View/FrequencyRelationships/GraphicalControl.xaml.cs
codymccoy: Good catch, i think we added that to help try to diagnose the radio button issue. I deleted it. A good reason to always put "todo: delete me".

View/FrequencyRelationships/GraphicalTableWithPlotControl.xaml
codymccoy: deleted

Brennan1994 (approved):

Brennan1994 (approved):

PR #452, opened 2022-07-20
Task review fixes

codymccoy:


UserControls/UserControls.vbproj
```diff
@@ -53,13 +53,13 @@
...
     </Reference>
     <Reference Include="fda-model, Version=1.1.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\USACE.HEC.FDA.Model.1.1.0-alpha-2022-07-15-2223\lib\net48\fda-model.dll</HintPath>
+      <HintPath>..\packages\USACE.HEC.FDA.Model.1.1.0-alpha-2022-07-19-1451\lib\net48\fda-model.dll</HintPath>
     </Reference>
     <Reference Include="Model, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\HEC.MVVMFramework.Model.1.0.0-alpha-2022.7.8.2331\lib\netstandard2.0\Model.dll</HintPath>
     </Reference>
     <Reference Include="Statistics, Version=1.1.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\USACE.HEC.FDA.Statistics.1.1.0-alpha-2022-07-15-2223\lib\net48\Statistics.dll</HintPath>
```
Brennan1994, line 62: Go ahead and drop this reference. This reference shouldn't exist at all. User controls doesn't need to know anything about statistics.

ViewModel/Alternatives/Results/DamageWithUncertaintyVM.cs
```diff
@@ -29,12 +29,12 @@ public DamageWithUncertaintyVM(AlternativeResults results, DamageMeasureYear dam
             if (double.IsNaN(discountRate))
             {
                 RateAndPeriodVisible = false;
-                ProbabilityExceedsValueLabel = "Probability that damage exceeds indicated value";
```
Brennan1994, line 32: For labels like these you were using a String Constants class for awhile. I've also seen in other people's code having string constants at the top of the class, just so they're in the same class they're used, but still easy to find and change easily. Does it make sense to use one of those strategies here?

Brennan1994: There's a few places you're using string labels in code in this pull request. I'll let this comment refer to  all of them.

Brennan1994 (changes requested):
Mostly looks good. just a couple notes

UserControls/UserControls.vbproj
Brennan1994: also doesn't know anything about model. Drop that guy too.

ViewModel/Alternatives/Results/DamageWithUncertaintyVM.cs
codymccoy: I agree, i will add them to the string constants class.

PR #529, opened 2022-09-07
Feature/actions releases

Brennan1994:


Brennan1994:
I ended up abandoning the github action to automagically upload the release asset. It was taking too much time to nail down. The only thing that's relevant in this pull request is that as apart of the experiment, I moved UserControls to dotnet 6, which needed to be done anyway.

PR #560, opened 2022-10-05
Feature/no terrain for wse

Brennan1994:


Brennan1994:
Cody and I reviewed and tested this manually

PR #562, opened 2022-10-05
Feature/sample from grids

Brennan1994:
This pull request contains both the additional feature of sampling from grids and cleaning up the model test libraries namespaces.  Review should be focused only on the first two commits. Don't waste your time reviewing all the files related to the namespace change. 

Currently this will not pass tests as the Sample Points functionality for a raster is not yet available in the RASMapper libraries that are currently under development. I'll coordinate with our friends over there to get that accomplished.


Brennan1994:
Libraries were updated. This is now fully implemented and functional

rnugent3 (approved):
Thank you!!

rnugent3:
By the way thanks for the heads up to only look at the first two commits

PR #564, opened 2022-10-06
Feature/correct wse for dry profiles

Brennan1994:


Brennan1994:
I almost forgot about this piece of stage damage. I wrote a method to make the adjustments to our depths in the same manner we did for the dean method. I have not incorporated it into the stage damage compute yet, just written and tested the logic. This will have to adjust each of our profiles as we use them.

rnugent3 (approved):
This is great. I am going to add this to my list of things to add to the release notes. We'll want to make sure we get buy-in on the barely dry adjustment.

PR #578, opened 2022-10-11
Computed stage damage

codymccoy:


Brennan1994 (approved):
We let this get way too big, but I think we're back in good graces agian. Merging.

PR #587, opened 2022-10-13
Feature/default data

rnugent3:
I've updated default data and made some changes/refactors along the way. The objective of the default data is having data that can be readily used for a compute. I think that will be a helpful pedagogical tool


rnugent3:
You might skip the big name refactor commit

Brennan1994 (approved):
Seems reasonable. I like the default data class.

PR #589, opened 2022-10-13
Plot error bounds on lognormal fix#582

Brennan1994:


rnugent3 (changes requested):
This looks great. One request for a tiny change - would you mind using "CL" in the line name instead of "CI" - I think we're looking at the confidence limit, not the confidence interval.

Brennan1994:
Done. 

> This looks great. One request for a tiny change - would you mind using "CL" in the line name instead of "CI" - I think we're looking at the confidence limit, not the confidence interval.

rnugent3 (approved):
Thanks, man.

PR #592, opened 2022-10-13
Fix bug in impact area importer

Brennan1994:


Brennan1994:
So easy I'm just gonna review it myself when the check passes.

PR #636, opened 2022-11-09
Structure details

rnugent3:
the model now produces a list of string on scenario stage damage that can be used to write structure details. 

I plan to implement the rest of the data writing using the same approach - passing around strings - except I do not anticipate that being as complicated as this where I need to add to existing strings, I think I'll just be able to add new strings. Let me know what you think of this approach?


fda-model/stageDamage/ImpactAreaStageDamage.cs
```diff
@@ -313,6 +319,96 @@ public void ReportMessage(object sender, MessageEventArgs e)
...
+            DepthsToStrings(deterministicInventory, ref structureDetails);
+            DamagesToStrings(deterministicInventory, STRUCTURE_DAMAGE, ref structureDetails);
+            DamagesToStrings(deterministicInventory, CONTENT_DAMAGE, ref structureDetails);
+            DamagesToStrings(deterministicInventory, OTHER_DAMAGE, ref structureDetails);
+            DamagesToStrings(deterministicInventory, VEHICLE_DAMAGE, ref structureDetails);
+
+            return structureDetails;
+        }
+
+        private void DamagesToStrings(DeterministicInventory deterministicInventory, string assetType, ref List<string> structureDetails)
```
Brennan1994, line 338: I'm a little surprised we don't have an enum for asset type. If so, that might be good to use, or good to generate unless we're purposely leaving this as a string. 

The string of If's could probably be cleaned up into a switch statement, but I think that's just sytlistic.

Brennan1994 (approved):

fda-model/stageDamage/ImpactAreaStageDamage.cs
rnugent3: We ended up with properties on asset type. This might be a nice refactor; we'd want to think about how to make use of the same enums in the model as in the view and view model.

PR #638, opened 2022-11-09
Feature/mapping multi field to same column

Brennan1994:


fda-model/structures/Inventory.cs
```diff
@@ -223,7 +223,7 @@ internal void GetUniqueDamageCatagories()
         _damageCategories = damageCatagories;
     }
 
-    private Inventory GetInventoryTrimmmedToPolygon(int impactAreaFID)
+    public Inventory GetInventoryTrimmmedToPolygon(int impactAreaFID)
```
rnugent3, line 226: are we confident that the impact area ID that we track in the model is the same as the impactAreaFID?

rnugent3: I remember getting myself confused on this, and I don't remember where we left off.

rnugent3 (approved):
Looks good

PR #647, opened 2022-11-14
test inv returns unique impact area ids and filters appropriately

Brennan1994:


Brennan1994:
Added a couple tests to verify the filter to polygon doesn't work. Seems like it does. I'm wondering if your shapefiles are in different projections.

rnugent3:
I will double-check, but I think that I put up that issue after I spent time playing with projections

PR #650, opened 2022-11-15
Feature/hydro profile interface

Brennan1994:


rnugent3:
@Brennan1994 looks like there is an invalid cast exception

Brennan1994:
You're right. I'm working on this.

Brennan1994:
@rnugent3 Updated. You can now create WSP in memory using the dummy profile object. Just construct it with the float[] you want back out from it. Let me know if you have any issues.

Issue #649, opened 2022-11-15
stage damage table with plot is off


rnugent3:
The stage-damage functions are now being returned with an array of stages and an array of histograms. Each histogram contains many observations and substantial variation. However, table with plot appears not to display that information - we're seeing weird numbers for the 5th and 95th percentiles.  

For this issue, I am going to use the Muncie example study data for the without-project conditions. In the below image, observe that we are looking at the stage-damage function for dam cat RES and asset cat CONTENT. There are 24 x values (stages) and 24 y values (histograms of damage). 

![image](https://user-images.githubusercontent.com/25163165/201792967-3a1c17be-e507-4e70-b13a-5089ecf3bfb0.png)

In the below image, observe that the min and the max of the last histogram are both big numbers (in the billions). This means that the 5th percentile and 95th percentile must be in the billions because the min and max are in the billions and the min and max reside outside the 5th and 95th percentiles. 

![image](https://user-images.githubusercontent.com/25163165/201793258-93a69254-a9b8-4d3c-84e1-fdd66ceca72a.png)

However, observe in the below image that the 5th and 95th percentiles are returned as 0, while the mean is positive and big. 

![image](https://user-images.githubusercontent.com/25163165/201793366-f7e78bf0-959c-4074-b929-69c371882425.png)

codymccoy:
This is where those values are getting created:
![image](https://user-images.githubusercontent.com/47757848/201967644-1a8d3bc8-c55e-4a6e-ad13-d67fee06659a.png)

Maybe something is wrong in the CDF function?

rnugent3:
Perfect. It should be the Inverse CDF function. Such an easy fix! Yay!

rnugent3:
This issue has been resolved

PR #651, opened 2022-11-15
Fixed bug #649 by switching to inverse cdf. Fixed bug where gridded h…

codymccoy:
…ydros were not validating correctly in the stage damage compute.


rnugent3 (approved):
Looks great thank you!

PR #653, opened 2022-11-15
Catagorizing all tests

Brennan1994:


Brennan1994:
@rnugent3 can you take a look at the failing unit tests? 

- HEC.FDA.ModelTest.unittests.StageDamageShould.StructureDetailsShould
- HEC.FDA.ModelTest.unittests.MessagingTests.ImpactAreaScenarioSimulationShould.ReportErrorsAndWarningsMessages

Brennan1994:
Also Missing the file from the Unit test for dbfreader ` string filepath = @"..\\..\\Resources\\ProbData.dbf";`

Brennan1994:
Ok. I've created a new category of test called Disk. Anything we run that requires a file path or connection to disk somewhere. These tests are the ones that are problematic. There's an issue when running these where two tests running in parrallel rely on the same file and lock eachother out. For this reason we probably need to only run these tests on our own machines where we can do it one at a time. Recatagorizing these effectively removes them from the automated build system. 

Unit tests run each pull request ~3min. Integration tests run once a week ~11min. Disk tests we should run on our own machines when we're working on those portions of the code.

rnugent3:
@Brennan1994 looks like you handled the messaging test? Structure details test is passing.

Brennan1994:
I did. There's some funny stuff happening with it we can talk about another time, but I think it only impacts the Listener class, which is just used for the test. 

I'm fightning github right now. For some reason I can't get tests to run on the github server, even though they're working on my machine.

Brennan1994:
getting close. [ImpactAreaScenarioSimulationShould.cs] might be the culprit.

rnugent3:
@Brennan1994 Thanks for figuring this out.

PR #663, opened 2022-11-18
Write stage damage details

codymccoy:


ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
```diff
@@ -51,13 +53,13 @@ public AggregatedStageDamageEditorVM(ChildElement elem, EditorActionManager acti
             if (IsManualRadioSelected)
             {
                 ManualVM = new ManualStageDamageVM(element.Curves);
-                CalculatedVM = new CalculatedStageDamageVM();
+                CalculatedVM = new CalculatedStageDamageVM(GetName);
```
rnugent3, line 56: I'm curious why GetName here does not need to be GetName()

rnugent3: - and in the other locations

rnugent3 (approved):
Looks good, aside from the GetName comment.

ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
codymccoy: Good question. I rarely write code like this, but this is passing in a method as a parameter. If you look at the constructor it takes a "Func<string>". This means that it requires a function that returns type string. If you used GetName() you would be passing in the string name and not the function. 

I imagine the next question is why am i doing this? We want to check that the user has entered a name when we go to write out the details file. This editor and UI has a parent panel that knows about the name, and then two sub panels, one for manual and one for computed. We need the computed panel to be able to ask what the name is everytime the compute button is clicked. So i pass in a GetName method to the child panel so that it can call that method on the parent panel.

codymccoy: I always have to look up the syntax when doing this sort of thing. Action<> is when you pass in a void and Func<> is when it has a return type.

rnugent3: Cool. Thanks for catching me up!

Brennan1994: This feels a little strange. Can we not just dissallow computing the curves if they don't have a name entered in the same way we disallow saving? I'm worried this is a more complex solution for very little gain. @codymccoy do you dissagree?

Brennan1994 (changes requested):
merge pending discussion on the GetName func.

ViewModel/AggregatedStageDamage/AggregatedStageDamageEditorVM.cs
codymccoy: The save button is in the parent panel so it can easily evaluate the name. The compute button is in the child panel so I needed to come up with a way for it to get access to the name. There are other methods possible, but I feel like passing the function into the child is the best approach in this situation.

codymccoy: Another approach that we do in java UI pretty regularly is to pass the parent panel into the children so that the children can access those methods. Since we only had one method I wanted access to I thought this was simpler, but if you prefer to pass the parent into the child in the ctor we could do that. Then in the child we could get name by doing parentPanel.Name

PR #676, opened 2022-12-01
now catch all cases of dry structures, not just null data.

Brennan1994:
This fix captures this case. Where a structure exists in a cell with wildly varying terrain elevations. Sampling through an HDF file, we will get the WSE of that cell when we sample the structure at that point, even through the terrain is above the wse. This won't happen with grids. 

image

This is a simple solution. It will apply our default correction even if we have data for the "dry" event. If a structure is dry, it's damage will be calculated at Terrain - 2ft, or Terrain -9ft, even if it's only "dry" by 1ft. 

image

In the second drawing. even though we can see the WSE at the structures is calculated at 4ft. (1ft below the surface) for the 5YR event, we will end up using (5ft -2ft = 3ft) for the 5YR event damages, because we determined the structure dry, and applied our fix. 

We could potentially do this more accurately by using the 4ft value. We'll have to add some complexity in tracking when to do that though.

### DIFF
--- a/fda-model/hydraulics/HydraulicDataset.cs
+++ b/fda-model/hydraulics/HydraulicDataset.cs
@@ -59,13 +59,13 @@ namespace HEC.FDA.Model.hydraulics
         {
             float offsetForDryStructures = 9;
             float offsetForBarelyDryStructures = 2;
-            float dryCellValue = -9999;
             if (nextProfileWses == null)
             {
                 for (int i = 0; i < wsesToCorrect.Length; i++)
                 {
+                    bool dryInCurrentProfile = wsesToCorrect[i] < groundElevs[i];
                     //The case where the largest profile has dry structures
-                    if (wsesToCorrect[i] == dryCellValue)
+                    if (dryInCurrentProfile)
                     {
                         wsesToCorrect[i] = (groundElevs[i] - offsetForDryStructures);
                     }
@@ -73,10 +73,12 @@ namespace HEC.FDA.Model.hydraulics
             }
             for (int i = 0; i < wsesToCorrect.Length; i++)
             {
-                if (wsesToCorrect[i] == dryCellValue)
+                bool dryInNextProfile = nextProfileWses[i] < groundElevs[i];
+                bool dryInCurrentProfile = wsesToCorrect[i] < groundElevs[i];
+                if (dryInCurrentProfile)
                 {
                     //The case where the next largest profile is also dry
-                    if (nextProfileWses[i] == dryCellValue)
+                    if (dryInNextProfile)
                     {
                         wsesToCorrect[i] = (groundElevs[i] - offsetForDryStructures);
                     }


rnugent3:
Beginning damage depth is going to be really important for structures with basements. If not in the structure inventory, I'm wondering if we should assume some default value, e.g. the negative of foundation height. 

@Brennan1994 you mentioned that a structure in a concrete jungle would not see water in its basement until we have positive depth. Beginning damage depth would be the way to make sure of that.

rnugent3:
I'm not exactly tracking why we'd need to do this "dry" adjustment if we'll always get back a WSE if there is data in that cell. 

RAS seems to be giving us a closer estimate than -2 and -9, no? 

I guess I'm asking why our adjustment to the hydraulics is better than the hydraulics as is?

Brennan1994:
> 

There are two cases where we'd have a dry structure as it relates to a RAS result. 

- 1.) We have the condition above, where a cell has a WSE elevation lower than the terrain of the structure, then it has one higher the next profile
- 2.) We have a cell that's totally dry , then wet above the structure terrain the next profile.

Our adjustment is tailored to the second condition. The current version on this branch treats both Case1 and 2 the same, which has a simplistic appeal. 


The alternative would be to treat them differently. Only apply our adjustment for case 1, and keep the RAS result for case 2. Yesterday I worried that would be difficult. Today I think it'd actually be quite easy. I'll put it into code quick, then I'll let you decide which to do.

Brennan1994:
Maybe for case 2 we test to see if ground -2 would create something non-monotonic, and if it would, we just set the WSE to GroundElev. 

I think that would be a reasonable solution.

Or, it's pretty attractive to just treat everything the same. If it's dry, it's Ground -2, or -9. If it's wet, it gets the result.

Brennan1994:
> Beginning damage depth is going to be really important for structures with basements. If not in the structure inventory, I'm wondering if we should assume some default value, e.g. the negative of foundation height.
> 
> @Brennan1994 you mentioned that a structure in a concrete jungle would not see water in its basement until we have positive depth. Beginning damage depth would be the way to make sure of that.


Understood. That seems like a reasonable solution. I'd suggest that I was maybe overthinking the problem. That's something that could be handled elsewhere in the code. This method is just meant to apply adjustments to the hydraulics.

Brennan1994:
> Maybe for case 2 we test to see if ground -2 would create something non-monotonic, and if it would, we just set the WSE to GroundElev.
> 
> I think that would be a reasonable solution.
> 
> Or, it's pretty attractive to just treat everything the same. If it's dry, it's Ground -2, or -9. If it's wet, it gets the result.

Bumping this comment. @rnugent3. How are you thinking we should move forward?

rnugent3:
I think it is attractive to treat everything the same. I trust your lead on this.

Brennan1994:
Perfect. We move forward with everything the same.

PR #685, opened 2022-12-06
Impact area structures

### LINKED ISSUE #679: deterministic value ratios
if the other or content-to-structure ratio uncertainty is deterministic, we still need a box in which to enter the ratio. No text box exists at the moment.

codymccoy:
Fixes #679

### DIFF
--- a/ViewModel/Inventory/OccupancyTypes/DeterministicControlVM.cs
+++ b/ViewModel/Inventory/OccupancyTypes/DeterministicControlVM.cs
@@ -1,0 +1,47 @@
+﻿using HEC.FDA.ViewModel.Utilities;
+using HEC.MVVMFramework.ViewModel.Implementations;
+using Statistics;
+using Statistics.Distributions;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace HEC.FDA.ViewModel.Inventory.OccupancyTypes
+{
+    public class DeterministicControlVM : ValidatingBaseViewModel, IValueUncertainty
+    {
+        private Deterministic _Deterministic;
+
+        public event EventHandler WasModified;
+
+        public double Value
+        {
+            get;
+            set;
+        }
+
+        public DeterministicControlVM():this(0)
+        {
+        }
+
+        public DeterministicControlVM(double value)
+        {
+            Value = value;
+            _Deterministic = new Deterministic(Value);
+        }
+
+
+        public ContinuousDistribution CreateOrdinate()
+        {
+            return _Deterministic;
+        }
+
+        public FdaValidationResult IsValid()
+        {
+            //todo: is this right?
+            return new FdaValidationResult();
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


ViewModel/Inventory/OccupancyTypes/DeterministicControlVM.cs
```diff
@@ -0,0 +1,55 @@
...
+        public ContinuousDistribution CreateOrdinate()
+        {
+            return _Deterministic;
+        }
+
+        public FdaValidationResult IsValid()
+        {
+            FdaValidationResult vr = new FdaValidationResult();
+            if(Value>10 || Value<0)
+            {
```
rnugent3, line 49: please add a comment somewhere that this validation is for the value ratios, which are currently the only object for which this class is used

ViewModel/Inventory/OccupancyTypes/ValueUncertaintyVM.cs
```diff
@@ -87,20 +93,14 @@ public ValueUncertaintyVM(ContinuousDistribution valueUncertaintyOrdinate)
...
         public abstract XElement ToXML();
   
 
         public ContinuousDistribution CreateOrdinate()
         {
-            //the currentVM can equal null. That is the deterministic case
+            //the currentVM can equal null. That is the non ratio deterministic case
             if (CurrentVM == null)
             {
                 return new Deterministic(0);
```
rnugent3, line 106: how is the returned deterministic distribution used?

ViewModel/Inventory/OccupancyTypes/ValueUncertaintyVM.cs
```diff
@@ -142,6 +142,11 @@ public void SelectionChanged(Object newValue)
             }
         }
 
+        /// <summary>
+        /// Used to raise an event that will put the "*" on the occtype and prompt the save message when closing.
```
rnugent3, line 146: thank you for the documentation

rnugent3 (approved):
Looks good. I've requested some documentation and for some info just for my awareness. Thank you!!

ViewModel/Inventory/OccupancyTypes/DeterministicControlVM.cs
codymccoy: I added comments to the top of the class as the entire class only applies to the deterministic ratio option.

ViewModel/Inventory/OccupancyTypes/ValueUncertaintyVM.cs
codymccoy: CreateOrdinate gets called when writing to xml. This happens when it gets saved and also to clone the object when opening for edit.
 It also get called when I translate the VM occtype objects into Model occtype objects for the stage damage compute. This is where there was a bug and I was not handling the new ratio value correctly. I'm glad you commented so that I could find that bug. Let me know if you want to know more. All of the occtype code is tough to slog through as there is so much going on.

rnugent3 (approved):

PR #688, opened 2022-12-07
Save checkbox state

codymccoy:


rnugent3:
it looks like changes to the way that histograms are being written/read back are causing the failing tests 

image

rnugent3:
@codymccoy I will review and address any other failing tests if necessary after you'd have the chance to work on #707

rnugent3 (approved):
Looks good! Thank you.

PR #697, opened 2022-12-09
Feature/stage damage histograms

rnugent3:


rnugent3:
@rtle76 my latest test demonstrates that we are calculating damage with uncertainty correctly, yet 2.0 stage damage functions do not match with 1.4.3 functions. I think this demonstrates that our stage damage functions are correct, and you can hold off spending further time on the muncie stage damage integration test. I think we'll probably need to spend some time working through some examples and documenting in the release notes so that we can advise the field on what to expect.

rtle76 (approved):

PR #713, opened 2022-12-21
Stage damage should tests

rtle76:
The ComputeDamageWithUncertaintyOneCoordinateShouldComputeCorrecly test in StageDamageShould.cs has been fixed


rtle76:
@rnugent3 I don't understand why this failed

rnugent3:
@rtle76 looks like ValueUncertaintyShould tests are failing. I bet that the terrible math I wrote into the code, I also wrote into my calculations of the expected values.

rnugent3 (approved):
great work, @rtle76

PR #743, opened 2023-01-26
Feature/plan names unsteady hydro

Brennan1994:


Brennan1994:
removes requirement for exactly 8 profiles. now loads Unsteady RAS files by plan title instead of file name. 

it looks to me to me like that swap could be done without too much harm because the path and probability arguments following it are what get used to construct the hydraulic profile. I'd like cody to take a look at it though to see if he sees any snags.

ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -166,36 +154,18 @@ public void FileSelected(string fullpath)
...
-                        prob += .1;
-                        AddRow(Path.GetFileName(dir), Path.GetFullPath(dir), prob);
-                    }
-                    //we might have some message for the user?
-                    if (!importResult.IsValid)
-                    {
-                        importResult.InsertMessage(0, "The selected directory contains 8 valid subdirectories and will ignore the following:\n");
-                        MessageBox.Show(importResult.ErrorMessage, "Valid Selection", MessageBoxButton.OK, MessageBoxImage.Information);
-                    }
+                    importResult.InsertMessage(0, "The selected directory contains 8 valid subdirectories and will ignore the following:\n");
```
codymccoy, line 167: Should this message still say 8? 
Maybe change to: The selected directory contains valid subdirectories and will ignore the following

ViewModel/Hydraulics/UnsteadyHDF/UnsteadyHDFImporterVM.cs
```diff
@@ -186,6 +186,12 @@ public void FileSelected(string fullpath)
             }
         }
 
+        private string getUnsteadyRASResultName(string file)
+        {
+            RasMapperLib.RASResults result = new RasMapperLib.RASResults(file);
+            return result.PlanAttributes.PlanTitle;
```
codymccoy, line 192: not sure what RasResults does here. If it can return null then we should probably add a null check here.

codymccoy (commented):
made some comments.

PR #760, opened 2023-02-06
Index points validation

codymccoy:


Brennan1994 (approved):
seems reasonable

Brennan1994:
image

TractableStageDamage failing because of an error in deterministic structure.

codymccoy:
The tractable stage damage tests are failing because of a bug described in #761. I don't think this has to do with my code, i think it has to do with a structure inventory that has fatal errors. I vote we merge this PR in to main.

rnugent3:
> The tractable stage damage tests are failing because of a bug described in #761. I don't think this has to do with my code, i think it has to do with a structure inventory that has fatal errors. I vote we merge this PR in to main.

I think I would rather fix the bugs and get the tests running before we go into main which I'll attempt to get done today. @Brennan1994 if you have other thoughts, I'll follow your lead.

Brennan1994:
Yeah, lets fix this before we get it in. I feel like we're being punished a little for letting this sit too long.

rnugent3:
The test is passing so I think we can merge. We still have a situation where we could end up passing null objects around. First step to fixing will be to address Issue #777.

rnugent3 (approved):

PR #763, opened 2023-02-07
remove bad validation logic. Add column mappings to match MuncieNSI. …

Brennan1994:
…fix 757


fda-model/structures/Inventory.cs
```diff
@@ -199,7 +199,7 @@ private void LoadStructuresFromSourceFiles()
                     ground_elv = GetRowValueForColumn<double>(row, _map.GroundElevCol, defaultMissingValue); //not gauranteed
                 }
                 double ff_elev = GetRowValueForColumn<double>(row, _map.FirstFloorElevCol, defaultMissingValue); // not gauranteed  
-                if ("".Equals(_map.FirstFloorElevCol) || row[_map.FirstFloorElevCol] == DBNull.Value)
+                if (ff_elev == defaultMissingValue)
```
rnugent3, line 202: This makes so much more sense to me than what was here before. Thank you!

rnugent3 (approved):
looks great, thanks @Brennan1994

PR #825, opened 2023-03-06
805 rename projects and folders

Brennan1994:


Brennan1994:
This is a bunch of file organization. Because we hodgepodged different projects together and didn't have a consistent naming scheme in the beginning our file structure hasn't matched out project/assembly names. This pull request goes through and makes those naming changes and puts projects in the right place. No substantial code has been added or removed.

PR #843, opened 2023-03-24
Bug/fragility

rnugent3:
solution as discussed - except I also took the stages from the fragility function to make sure that we keep all of the information in both functions

### DIFF
--- a/HEC.FDA.Model/paireddata/PairedData.cs
+++ b/HEC.FDA.Model/paireddata/PairedData.cs
@@ -247,7 +247,24 @@ namespace HEC.FDA.Model.paireddata
                 newXvals.Add(stageFromStageDamage);
                 newYvals.Add(probabilityWeightedDamage);
             }
-            return new PairedData(newXvals.ToArray(), newYvals.ToArray());
+            for (int i = 0; i < newSystemResponse.Xvals.Length; i++)
+            {
+                double fragilityStage = newSystemResponse.Xvals[i];
+                bool fragilityStageIsInStages = newXvals.Contains(fragilityStage);
+                if (!fragilityStageIsInStages)
+                {
+                    double probabilityOfFailure = newSystemResponse.Yvals[i];
+                    double unweightedDamage = f(fragilityStage);
+                    double probabilityWeightedDamage = probabilityOfFailure*unweightedDamage;
+                    newXvals.Add(fragilityStage);
+                    newYvals.Add(probabilityWeightedDamage);
+                }
+            }
+            double[] stages = newXvals.ToArray();
+            double[] damages = newYvals.ToArray();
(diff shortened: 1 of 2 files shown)


HEC.FDA.ModelTest/unittests/PairedDataShould.cs
```diff
@@ -89,14 +89,19 @@ public void Multiply()
         {
             double[] multiplierXs = { 2, 3, 4 };
             double[] multiplierYs = { .5, .5, .5 };
-            double[] expectedXs = { 1, 1.999, 2, 3, 4, 4.001, 5 };
-            double[] expectedYs = { 0, 0, 1, 1.5, 2, 4.001, 5 };
+
+            double[] expectedXs = { 1, 2, 3, 4, 5 };
+            double[] expectedYs = { 0, 1, 1.5, 2, 5 };
             PairedData multiplier = new PairedData(multiplierXs, multiplierYs);
```
Brennan1994, line 95: I'd like to see 3 test cases: one where the multiplier has fewer points than the base, one where it has more, and one where the coordinate count matches.

Brennan1994 (changes requested):

HEC.FDA.Model/paireddata/PairedData.cs
```diff
@@ -232,62 +232,69 @@ public double integrate()
...
-                        break;
-                    }
-                }
-            }
-            // calculate damages for the range of the fragility curve
-            for (int index = 0; index < systemResponseFunction.Xvals.Count(); index++)
+            IPairedData newSystemResponse = EnsureBottomAndTopHaveCorrectProbabilities(systemResponseFunction);
+
+            List<double> newXvals = new List<double>(); //xvals are stages in the stage-damage function
+            List<double> newYvals = new List<double>(); //yvals are damage*prob(failure)
```
Brennan1994, line 238: This is probably fine, but arrays are going to be more efficient than lists. If we find this to be slow when we do profiling, we may have to come back and refactor. List seem elegant to me for this solution so I'd leave em for now.

Brennan1994 (changes requested):

HEC.FDA.ModelTest/unittests/PairedDataShould.cs
rnugent3: great idea

Brennan1994:
@rnugent3 waiting on that unit test update before I approve. Just bumping this to make sure you know I'm tracking this as in your court.

rnugent3:
Thanks for the ping, Brennan.

rnugent3:
@Brennan1994 this is ready for backcheck.

Brennan1994 (approved):

PR #849, opened 2023-04-01
828 projection needs to be set at the study level

Brennan1994:
This sets up the MVP UI for projections. You are now able to specify a prj file to be the study projection. Upon saving the study properties, it will be copied into the study directory under the Projection folder. If a file already exists there, it will automattically populate as the project projection file. 

This has not been tied into the compute yet. That comes next, but I want to get this into main because it's a nice chunk of work and thought.


Brennan1994:
Organizationally it makes sense to keep projection with other study properties, so that's it's home right now.    We'll need to communicate with our workflow/descriptions/menu items that it needs to match the projections used in the hydraulic model. Thinking about maybe adding a button in the hydraulic importers that opens up the study properties window.

PR #851, opened 2023-04-03
828 projection needs to be set at the study level

Brennan1994:
This loops in the study projection to the stage damage compute. We now look for the projection file when we go to recompute, and if it's there, we reproject. If we don't, we just don't  reproject. 

We no longer pull any projection information from the terrain.


rnugent3 (approved):
looks good thank you

Issue #768, opened 2023-02-09
Recreate single coordinate stage damage tests


rnugent3:
Should we go ahead and write that test before we merge?

_Originally posted by @Brennan1994 in https://github.com/HydrologicEngineeringCenter/HEC-FDA/pull/752#discussion_r1099485495_

rnugent3:
The compute damage at one coordinate with uncertainty unit tests have been deleted from the test suite to accommodate the stage-damage algorithm refactor. Please re-write the tests in a way that allows us to leverage the existing test data generated for these tests.

rnugent3:
24hrs

PR #857, opened 2023-04-05
wrote stage damage with uncertainty tests

rnugent3:
I have written a test that has two purposes:

1. Replace the damage with uncertainty tests on stage damage
2. Serve as test for profiling stage-damage compute 

I attempted a different way to make sure that test values are within a reasonable tolerance. Doing the assertion for multiple cases that have different orders of magnitude is complicated. If you see a better way, let's do it.


rnugent3:
This PR Fixes #768. I forgot to include that in the commit message.

rnugent3:
The good news is that we've confirmed that damage with uncertainty in 2.0 reasonably matches damage with uncertainty in 1.4.3 given that the function is calculated using a graphical stage-frequency function.

rnugent3:
When we refactor stage-damage as planned, this test will assure that we do not mess up uncertainty bounds - which has been our concern about sampling the set of occupancy types once for a given stage-damage realization.

rnugent3:
all of my tests appear to pass but the test run is being aborted

rnugent3:
failure message: "vstest.console process failed to connect to testhost process after 90 seconds. This may occur due to machine slowness, please set environment variable VSTEST_CONNECTION_TIMEOUT to increase timeout."

rnugent3:
tests are now passing but I will leave my comments above for reference

Brennan1994:
Was there an issue with your tests that caused the abortion that you fixed or was it just random?

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
```diff
@@ -0,0 +1,289 @@
...
+        private static CurveMetaData stageFreqMetaData = new CurveMetaData("probability", "stages", "graphical stage frequency");
+        private static GraphicalUncertainPairedData frequencyStage = new GraphicalUncertainPairedData(probabilities, stages, equivalentRecordLength, stageFreqMetaData, true);
+
+        //Hydraulics for a given impact area 
+        private static HydraulicDataset ComputeHydraulicDataset()
+        {
+            List<IHydraulicProfile> dummyHydraulicProfiles = new List<IHydraulicProfile>();
+            for (int i = 0; i < stages.Length; i++)
+            {
+                float[] profileStages = new float[quantityProfilesAndStructuresPerImpactArea];
```
Brennan1994, line 54: I dont like this name. maybe make two static ints. ProfilesPerImpactArea and StructuresPerImpactArea. Those two things don't inherently match and I, finding it confusing reading it.

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
```diff
@@ -0,0 +1,289 @@
...
+                {
+                    profileStages[j] = (float)stages[i];
+                }
+                DummyHydraulicProfile dummyHydraulicProfile = new DummyHydraulicProfile(profileStages, probabilities[i]);
+                dummyHydraulicProfiles.Add(dummyHydraulicProfile);
+            }
+            HydraulicDataset hydraulicDataset = new HydraulicDataset(dummyHydraulicProfiles, Model.hydraulics.enums.HydraulicDataSource.WSEGrid);
+            return hydraulicDataset;
+        }
+        private static HydraulicDataset hydraulicDataset_34 = ComputeHydraulicDataset();
```
Brennan1994, line 65: What is the 34 and 56?

Brennan1994: rescinded. I see. Maybe call them impactAreaA and impactAreaB though? not sure I get the significance of these numbers.

rnugent3:
I think it was random

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
```diff
@@ -0,0 +1,289 @@
...
+        #endregion
+
+        #region Other objects 
+        private static ConvergenceCriteria convergenceCriteria = new ConvergenceCriteria(minIterations: 100, maxIterations: 200000);
+        RandomProvider randomProvider = new RandomProvider(seed: 1234);
+        #endregion
+
+        #region Expected Values
+        //The modeling used to generate the expected values can be found at the below link. 
+        //https://drive.google.com/file/d/17IvI01PBXa0C97dlBOKLn0Z0odGTJ663/view?usp=share_link
```
Brennan1994, line 148: You'll never get the credit you deserve for doing this. but it's pretty freakin' cool.

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
```diff
@@ -0,0 +1,289 @@
...
+            for (int i = 0; i < stageAtWhichToCheckForDamage.Length; i++)
+            {
+                expected_conf95_damageDists_34[i] = new Normal(expected_mean_residentialDamage_34[i], expected_standardDeviation_residentialDamage_34[i]).InverseCDF(conf95Prob);
+                expected_conf95_damageDists_56[i] = new Normal(expected_mean_residentialDamage_56[i], expected_standardDeviation_residentialDamage_56[i]).InverseCDF(conf95Prob);
+            }
+
+            for (int i = 0; i < stageAtWhichToCheckForDamage.Length; i++)
+            {
+                //TODO: These two lines are being commented out. They do not pass at this time. 
+                //Making sure that the test passes for Impact Area 34 will take place during the stage-damage refactor
```
Brennan1994, line 239: should these comments be removed? I don't see the commented out lines.

Brennan1994 (approved):

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
rnugent3: the two things have to match because they're tied together by index

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
rnugent3: done

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
rnugent3: thanks man

HEC.FDA.ModelTest/integrationtests/UncertainStageDamageTests.cs
rnugent3: done

PR #858, opened 2023-04-06
852 analytical frequency curve should plot on normal probability axis

Brennan1994:


Brennan1994:
This does not fully address the issue of confidence limits yet. It applies them in one location, and adds the model logic to compute them.

HEC.FDA.ModelTest/unittests/extensions/ContinuousDistributionExtensionsShould.cs
```diff
@@ -0,0 +1,47 @@
...
+
+
+        [Fact]
+        //https://www.hec.usace.army.mil/confluence/sspdocs/ssptutorialsguides/expected-moments-algorithm-study-guide/user-defined-functions-for-lpiii-and-gev-frequency-curves#UserDefinedFunctionsforLPIIIandGEVFrequencyCurves-Exercise
```
rnugent3, line 20: cool!

rnugent3:
I ran into an uncaught exception 

image

rnugent3:
The exception happens when I attempt to view a graphical frequency function.

rnugent3 (changes requested):
Please troubleshoot the overflow exception.

Brennan1994:
The latest update updates the tracker to use the probability rather than the Z score, so you can check exact probabilities anywhere on the lines. 

I was unable to reproduce your error. Please send me your dataset if that's still an issue.

rnugent3 (approved):
terrific

PR #865, opened 2023-04-10
Added regulated-unregulated elements to the computed stage damage table.

codymccoy:


Brennan1994:
I've reviewed this. UI looks good and code looks reasonable, but i don't have a study I can compute. Always hit the memory issue on stage damage. 

Richard, could you pull and give this a run?

codymccoy:
I pushed my code. Ready to test. I can't say that the actual compute is working as it should but I can verify that the curves look different when using a reg-unreg curve:
Without:
image

With:
image

My compute time with Muncie was about 25 seconds.

rnugent3 (approved):
Looks good, thank you.

PR #876, opened 2023-04-26
Remove AED from Alternative Results Report

### LINKED ISSUE #823: focus damage reporting
We should remove EAD reporting at the alternative level.

- Scenario damage is EAD
- Alternative damage is AAED
- Alternative comparison damage is EAD Reduced and AAED Reduced

Brennan1994:
image


rnugent3 (approved):
Looks great, thank you.

PR #878, opened 2023-04-29
Bug/randomness

### LINKED ISSUE #798: uncontrolled randomness
The results from one simulation are slightly different if re-computed. Did we miss a seed somewhere? 

@HenryGeorgist if you have a couple minutes, I'd appreciate your help investigating. The means are the same but the distributions are not. The quartiles are off by less than $100 - a very small fraction. I am not sure if it matters, but it looks like the difference is increasing with increasing cumulative probability.  

My greater concern for **this** is that I think the issue is related to #782. 

## Existing Condition Future Year Left Bank EAD Results
image

## Same Results Recomputed 
image

rnugent3:
This Fixes #798.


HEC.FDA.Model/compute/ImpactAreaScenarioSimulation.cs
```diff
@@ -18,6 +18,7 @@
 using System.Collections.Concurrent;
 using Statistics.Distributions;
 using HEC.FDA.Model.extensions;
+using Microsoft.VisualBasic;
```
rnugent3, line 21: I am not sure why this was added. Is it a problem?

rnugent3:
I think it would be possible to obtain performance improvements (reduction in memory requirements) by chopping up min iterations into 10 or 20 chunks.

rnugent3:
Stage damage tests are failing due to memory issues

Brennan1994 (commented):
I might need this explained to me rather than giving it an independent review. I'm not familiar enough with the over class and problem to provide much useful feedback.

rnugent3:
Last night, I dreamed up an alternative solution that would not require sorting. I'm going to make some changes tomorrow or over the weekend, and then let's talk.

rnugent3:
the compute appears to be pretty fast - in 40 seconds we can run 2 million iterations.

I need to add back in the histogram for plotting assurance of aep

rnugent3:
tests failing on remote but passing locally are my favorite tests

rnugent3:
the tests test more than the code............

HEC.FDA.Model/structures/OccupancyType.cs
```diff
@@ -201,6 +201,13 @@ internal OccupancyTypeBuilder(OccupancyType occupancyType)
             public OccupancyType build()
             {
                 _occupancyType.Validate();
+                //unregister?
+                //unregister each of the objects that got built within the builder pattern 
+                //reflect on an instance of this, ask for any field that is a validation data error logger thingy
+                //on that thing, if not null, unregister 
+
+                //need something like IUnregisterFromMessageHub
+                //in computational method, at the very beginning, we call Validate()
```
rnugent3, line 210: Kill commented code

HEC.FDA.Statistics/Distributions/IDistributionFactory.cs
```diff
@@ -38,13 +38,19 @@ public static string PrintParamaterizationRequirements(IDistributionEnum type)
...
             }
         }
-
+        /// <summary>
+        /// we have to keep this because of IDistribution
+        /// </summary>
+        /// <param name="sample"></param>
+        /// <param name="nBins"></param>
+        /// <returns></returns>
         internal static Statistics.Histograms.Histogram Fit(IEnumerable<double> sample, int nBins)
```
rnugent3, line 47: IDistribution has a bunch of baggage. That interface needs broken up into multiple interfaces so classes don't have to implement bogus methods.

rnugent3: Also, Delete IDistribution Factory and replace it with the proper constructors all over the place. It's only used in tests .

HEC.FDA.Statistics/Histograms/ThreadsafeInlineHistogram.cs
```diff
@@ -195,6 +195,7 @@ public ThreadsafeInlineHistogram()
                 AddObservationToHistogram(0,i);
             }
         }
+
```
rnugent3, line 198: Delete ThreadsafeInlineHistograms

PR #880, opened 2023-05-05
822 batch computes

codymccoy:


rnugent3 (changes requested):
This work is phenomenal. You identified several requirements and took the initiative to solve them. I really appreciate the initiative. I've found at least one bug. I'll document below.

rnugent3:
I opened the attached study and computed all scenarios. Upon completion of the compute, we had for all scenarios but one, green check marks. There is a yellow warning sign on the Future Levee scenario that indicates that the scenario has been edited since last compute - but I did not edit. 

[WS6Model_2.zip](https://github.com/HydrologicEngineeringCenter/HEC-FDA/files/11467842/WS6Model_2.zip)

image

rnugent3:
We are missing Long-term Exceedance Probability in the performance summary. That might have been my fault. Please add a set of Long-Term Exceedance Probability columns for 10, 30, and 50 years.

rnugent3:
It appears that the results displayed in the performance summary are incorrect. 

image

Below are examples of the calls you'll need: 

```
            double meanAnnualExceedanceProbability = baseScenarioResults.MeanAEP(impactAreaID, thresholdID);
            double medianAnnualExceedanceProbability = baseScenarioResults.MedianAEP(impactAreaID, thresholdID);

            double exceedanceProbability = 0.1;
            double assuranceOfThreshold = baseScenarioResults.AssuranceOfEvent(impactAreaID, 1 - exceedanceProbability, thresholdID);

            int years = 10;
            double longTermExceedanceProbability = baseScenarioResults.LongTermExceedanceProbability(impactAreaID, years, thresholdID);

            double assuranceOfAEP = baseScenarioResults.AssuranceOfAEP(impactAreaID, exceedanceProbability, thresholdID);
```

rnugent3:
If it would not take much time, please add an Assurance of AEP table to the summary report. This can take place in a third table, with the scenario information and performance parameter columns put in that third table, too.

rnugent3:
Please round the AAED values to two decimal places. 

image

rnugent3:
Please move the summary option from the report combo box to the top of the With-Project Alternative project. 

image

rnugent3:
It appears that the analysis year combo box is not plugged in. The results appear to be the same for both analysis years, but this is not the case. 
image

codymccoy:
I started by adding a 3 second buffer on the check between the compute date and the edit date but then i realized that that doesn't solve the issue of two different dates showing in the tooltip (only off by 1 second, but still). I couldn't come up with a super clean way of doing it, but I know make sure that the exact same date is used for both the edit and compute date when a compute is done.
image

codymccoy:
I added the long term risk

> It appears that the results displayed in the performance summary are incorrect.
> 
> image
> 
> Below are examples of the calls you'll need:
> 
> ```
>             double meanAnnualExceedanceProbability = baseScenarioResults.MeanAEP(impactAreaID, thresholdID);
>             double medianAnnualExceedanceProbability = baseScenarioResults.MedianAEP(impactAreaID, thresholdID);
> 
>             double exceedanceProbability = 0.1;
>             double assuranceOfThreshold = baseScenarioResults.AssuranceOfEvent(impactAreaID, 1 - exceedanceProbability, thresholdID);
> 
>             int years = 10;
>             double longTermExceedanceProbability = baseScenarioResults.LongTermExceedanceProbability(impactAreaID, years, thresholdID);
> 
>             double assuranceOfAEP = baseScenarioResults.AssuranceOfAEP(impactAreaID, exceedanceProbability, thresholdID);
> ```

@rnugent3 The "impactAreaID" in the above calls are for each individual impact area within the impact area set. That would change the rows to be a row for each specific impact area. Is that what we want?

rnugent3:
We do need results for each specific impact area - I must have left that out of the table dimensions?

rnugent3:
Ideally we do scenario level and impact area - scenario level for each set of results. Would that be tough?

codymccoy:
> Please round the AAED values to two decimal places.
> 
> image

I now round to two decimal places. I also gave the data grid a min width so that the header doesn't get clipped. This means that the table has some dead space on the right side if there aren't enough dam cats to fill the min width but I think that is better than having the header clipped.
image

codymccoy:
> Please move the summary option from the report combo box to the top of the With-Project Alternative project.
> 
> image

I added the summary option to the top combo. When that option is selected i show both summary tables (EAD and AAEQ) and I remove the other combo options. I think this is cleaner, but let me know if you want a different design.
image

codymccoy:
Looks like i missed that there are two ead tables. One for base year and one for future year. I will add that now.

codymccoy:
@rnugent3 Updated summary UI. Does the third table need a label above it? 
image

codymccoy:
> It appears that the analysis year combo box is not plugged in. The results appear to be the same for both analysis years, but this is not the case. image

Maybe this behavior was only for the summary option? i am not seeing it on the other "report" options and I have moved the summary option to the higher level so hopefully that solves this issue.

codymccoy:
Updated labels above the tables in the alt comp report summary:
image

codymccoy:
@rnugent3 All of the above issues have been addressed. This PR is now ready for your review again.

rnugent3:
damage figures do not match between the summary results and the individual scenario results, see screen shots below. performance statistics do match. 

image
image
image

rnugent3:
please add a space between EAD and the analysis year 

image

rnugent3:
@codymccoy I have finished my review - one small bug and one tiny formatting change and we'll be ready to get this merged....THANK YOU

rnugent3 (approved):

PR #889, opened 2023-05-18
remove message hub from histograms.

Brennan1994:
des perez study now opens in ~3 minutes. Memory no longer the big issue. Further gains should come from parallelizing the bootstrap. which I'll work on next.


rnugent3 (approved):
looks good thanks man

PR #897, opened 2023-06-01
Scenario workflow 818

codymccoy:
There are several failing unit tests. I can't tell if they are from my work on not. I checked out main and it had several of the same failing tests.


Brennan1994:
I found this too. I think those test that are also failing on main are related to some work Richard's been doing. I don't understnad how they passed on main on the build server. I'lll review this shortly.

rnugent3:
I don't think I can be of much help this week - I'm without a working computer (it got a little wet). 

Let's jump on a call if the failure is not straightforward.

Brennan1994 (approved):

PR #902, opened 2023-06-07
Scenario threshold 820

codymccoy:
Several github issues completed in this PR.


Brennan1994:
Was this known to break backward compatability? I'm getting tons of exceptions loading old studies. I wrapped the following code in a try catch

Brennan1994 (approved):

rnugent3 (changes requested):
Stellar work!! I have a few requests for changes - some small UI stuff. 
image
image
image

codymccoy:
Threshold stage label updated:
image

I added a min width to these combos so that they look better when nothing is in them:
image

codymccoy:
The value gets set when the lateral structure is selected and goes back to zero when the blank option is selected. 
image

When there is a lateral structure then the checkbox is disabled:
image

PR #909, opened 2023-06-22
fix #905

### LINKED ISSUE #905: edits to hydraulics importer
we need to change the default probabilities and return intervals to all zeros so it is obvious to the user that edits need to be made 

we also need to add a tool tip to the save button on the hydraulics importer to remind the user to set their recurrence intervals

this will be an issue for the user - it happened during the workshop that a user did not adjust their probabilities

Brennan1994:
image

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -27,7 +27,6 @@ jobs:
       
     - name: Add Github and Nexus Source
       run:  |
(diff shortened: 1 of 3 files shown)


Brennan1994:
default editor now loads with obviously bad values. Throws error if you try to continue without changing them

codymccoy (approved):
Looks good.

PR #912, opened 2023-07-05
Bug/fix warnings

Brennan1994:


rnugent3 (approved):
really great work, thanks for figuring out a better way to serialize

PR #918, opened 2023-07-25
Overlapping curves now just checks that the lp3 is contained by the r…

Brennan1994:
…ating curve

Addressing this problem: 

image

rating curve covered the entirety of the specified flow frequency, but still checked out as not overlapping because the mins and maxs are so different.


rnugent3:
Couple of questions, @Brennan1994.

Should we also check if a graphical flow matches up with a rating curve? 

Does it not matter if a stage-freq does not match up with a stage-damage? 

Thanks for taking a look at this.

Brennan1994:
I've refactored this to be a little easier to understand I think. There's no more branching logic. Just checking each possible combination. I'd like to get your feedback on what the actual requirements are for each of these curve combinations. Do they need to Match. or do they need to be contained by each other? Like, a system response curve might only be defined to top of levee, and that seems ok as long as that was accounted for in the model logic. 

The original logic, I think, was requiring everything to match within some tolerance, but I don't think that's what we actually need. I'd like some input on each of these. Each curve comparison is broken into it's own method so we can customize them however we like. I'm just not familar enough with the compute to make these assertions without investing a lot going back and forth between compute and here.

rnugent3:
I don't think that we need to check for overlap anymore as part of the compute validation. We could check to advise the user. 

Curve overlap validation was a way to address this issue: https://github.com/HydrologicEngineeringCenter/fda-model/issues/290

The issue occurred in ThreadsafeInlineHistogram - but we're not using Threadsafe any longer. 

I have done some testing with some wildly non-overlapping curves, it appears that Histogram can handle flat/zero distributions just fine. 

I'd recommend removing curve overlap as required for validation to something more like a warning level.

rnugent3:
We could pursue warning the user about curve overlap for a later version, to boot

Brennan1994:
Ultimately we decided not to ditch the curve overlapping stuff. This is now just a bunch of cleanup. refactor, formatting changes, and a fix for a broken unit test.

HEC.FDA.Model/compute/ImpactAreaScenarioSimulation.cs
```diff
@@ -93,10 +94,9 @@ public ImpactAreaScenarioResults Compute(IProvideRandomNumbers randomProvider, C
         public ImpactAreaScenarioResults Compute(IProvideRandomNumbers randomProvider, ConvergenceCriteria convergenceCriteria, CancellationToken cancellationToken,
             bool computeDefaultThreshold = true, bool giveMeADamageFrequency = false, bool computeIsDeterministic = false)
         {
-            //Validate();
-            if (!CanCompute(convergenceCriteria, randomProvider))
+            if (!CanCompute(convergenceCriteria))
             {
-                _ImpactAreaScenarioResults = new ImpactAreaScenarioResults(_ImpactAreaID, true);
+                _ImpactAreaScenarioResults = new ImpactAreaScenarioResults(_ImpactAreaID, true); //I would like to just return regular Null here but I'm unsure who is relying on this behavior. BBB
                 return _ImpactAreaScenarioResults;
```
rnugent3, line 100: we don't have the logic in place to abort if can't compute

rnugent3 (approved):
terrific, thank you!

HEC.FDA.Model/compute/ImpactAreaScenarioSimulation.cs
rnugent3: refactoring to stop passing around null objects would be a solid reduction in technical debt

PR #919, opened 2023-07-27
Dynamic Tab Unique Names

Brennan1994:
we found we couldn't open a new tab for stage damage if there was already one open. I think the issue was lack of a unique name , which is the third argument in the dynamic tab constructor. 

I also just cleaned up some formatting. I really only want your confirmation that my fix makes sense to you with the dynamic tabs. IT seems to work in the UI.


codymccoy (approved):
Looks good to me. Yes, that is the point of the unique name. The tab controller looks to see if something with that unique name already exists, if it does, it selects that tab in the tab control, if it doesn't, then it opens up another one. I don't see a problem with using the date time. The unique name never shows to the user.

PR #924, opened 2023-07-31
357 update convergence criteria

rnugent3:
This is a small change. We've played with convergence criteria for some time, but we're at the place now where it should just match 1.4.3. I think we'll have the option to revisit later. 

Notice that convergence is different for EAD as it is for stage-damage. This is how 1.4.3 is wired. 

I worked on some tech debt along the way - better aligning responsibilities, though still not perfect.


rnugent3:
some weird changes came through in the test files, I am going to do a little more housecleaning

rnugent3:
I have identified the issue in the failing test, resolving shortly

rnugent3:
@Brennan1994 this is ready for your review. Thank you.

Brennan1994 (approved):

PR #933, opened 2023-08-14
877 occupancy type bug content value uncertainty entry isnt functioning properly

Brennan1994:


Brennan1994:
This stuff is kind of a mess. Need to design a decent test to prove this works. I'm still not convinced.

rnugent3 (changes requested):
I am still seeing the situation where the occ type as entered into the UI does not show up the same in the model. In the below screen shots, you'll see that the occupancy type has `true` for `ComputeOtherDamage` and `ComputeVehicleDamage` despite none of the occupancy types having data on those asset types. You can also see that there is a `_VehicleValueError` object which should not exist. 

image
image

My guess is that fake objects are being constructed in the case that the object has not been specified by the user to support the occ type builder pattern in a "simple" or "clean" way that has messed up our flagging system. When a vehicle value error object is added to the occupancy type as part of the builder pattern, we switch `ComputeVehicleDamage` to `true`

rnugent3:
I think that the common occ type build is happening here:
https://github.com/HydrologicEngineeringCenter/HEC-FDA/blob/39956194ccbd708986e79571cfe9a8f8e724ecb1/HEC.FDA.ViewModel/Inventory/InventoryElement.cs#L305

so we either need a way to build an occupancy type in a way that matches the scope of the user input data or revise our strategy for identifying whether to compute damage for a given asset type.

rnugent3:
When I click in to edit the occupancy types being used in a given stage-damage compute, I find that the selection for "By Content Value" and "By Content-To-Structure Value Ratio" has not been saved. 


When I make the correct selection, save the edits to the occ type, and recompute stage-damage, the correct selection fails to show up in the model. 

image

image

Brennan1994:
I definitely introduced that visual bug where neither radio button is selected. I will fix that first. We should chat about the remaining disagreement between view and model though. I'm not sure I'm seeing that on my end.

Brennan1994:
ok. There's a lot of kreft cleanup going on in this PR. The only thing that really matters is the CreateModelOcctype method. That's where we're converting from vm to model world. 

I think I've proven throught he UI this works., double check me. n

rnugent3 (approved):
looks great thank you!!

PR #934, opened 2023-08-17
926 compute alternatives doesnt appear to be working

Brennan1994:


Brennan1994:
The linked issue was a result of an analysis period that didn't cover the span of the analysis years. The saving issue was part of that. This PR does not address messaging at all.

rnugent3 (approved):
looks good, thanks, Brennan!!

PR #944, opened 2023-08-29
Bug/occtypes

rnugent3:
some bug fixes and some cleanup


HEC.FDA.Importer/AsciiImportExport/AsciiImport.cs
```diff
@@ -855,6 +856,14 @@ void ProcessOcctypeFields()
 
             }
         }
+
+        private void FixRatioParameters()
```
Brennan1994, line 860: can i get a comment to describe what this function is doing? It's unclear to me.

Brennan1994 (changes requested):

HEC.FDA.Importer/AsciiImportExport/AsciiImport.cs
rnugent3: sure thing!

Brennan1994 (approved):

PR #953, opened 2023-09-06
Bugs/sdo

rnugent3:


rnugent3:
image

HEC.FDA.View/Study/StudyView.xaml.cs
```diff
@@ -92,7 +92,7 @@ private void UserControl_Loaded(object sender, RoutedEventArgs e)
 
         private void QuickStartGuide_Click(object sender, RoutedEventArgs e)
         {
-            string quickStartLink = "https://www.hec.usace.army.mil/confluence/display/FDAQSG/HEC-FDA+Quick+Start+Guide";
+            string quickStartLink = "https://www.hec.usace.army.mil/confluence/fdadocs/fdaqsg/hec-fda-quick-start-guides-152636942.html";
```
HenryGeorgist, line 95: this should not be how you do this. use a FWLINK ID so you can maintain and never get page not found.

HEC.FDA.Model/stageDamage/ScenarioStageDamage.cs
```diff
@@ -2,6 +2,7 @@
 using Statistics;
 using HEC.FDA.Model.paireddata;
 using HEC.FDA.Model.interfaces;
+using Microsoft.VisualBasic;
```
HenryGeorgist, line 5: Why is the reference to VisualBasic being added? is it because of the way you are formatting this string?

HEC.FDA.Model/structures/Structure.cs
```diff
@@ -107,7 +99,7 @@ public ConsequenceResult ComputeDamage(float waterSurfaceElevation, List<Determi
             if (analysisYear > YearInService)
             {
                 //Beginning damage depth is relative to the first floor elevation and so a beginning damage depth of -1 means that damage begins 1 foot below the first floor elevation
-
+                //if not defined by the user, the beginning damage depth is equal to the negative of foundation height
```
HenryGeorgist, line 102: is this standard? i guess it is a decision, but typically i assume beginning damage is foundation height not ground.

HenryGeorgist (changes requested):

Brennan1994 (changes requested):

HEC.FDA.View/Study/StudyView.xaml.cs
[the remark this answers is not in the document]
rnugent3: I've planned to figure out the forward link stuff, some time this week.....

HEC.FDA.Model/structures/Structure.cs
rnugent3: it is common in depth-percent damage functions for percent damage to be positive below the first floor - I think it we assumed beginning damage was foundation height/FFE, we'd truncate that damage, right?

HEC.FDA.Model/stageDamage/ScenarioStageDamage.cs
rnugent3: it was the way I was getting the date and time, I changed the date time property to use System

Brennan1994 (approved):

Brennan1994:
I'm merging without Wills review being satisfied because I want to get the rest of this in ASAP. Richard has made a new issue to address Our complaints about forward link.

Issue #963, opened 2023-09-20
Changes to Scenarios Editor for Total Risk Calculation


rnugent3:
Let's use this issue to track the view changes for the total risk calculations. I have updated the screen shots to remove the possibility of using an exterior-interior function if modeling nonfail risk + fail risk. 

![image](https://github.com/HydrologicEngineeringCenter/HEC-FDA/assets/25163165/f4c42b53-0a29-4337-91b8-912aa1269e1e)
![image](https://github.com/HydrologicEngineeringCenter/HEC-FDA/assets/25163165/4c52a4fa-5d5b-4d9a-95a3-4ca4d0870c21)

rnugent3:
@codymccoy please disregard my note about not allowing use of the default system response. 

default system response is OK - total probability still applies

codymccoy:
@rnugent3 How does this look? I don't have the checkbox logic plugged in yet. 
![image](https://github.com/HydrologicEngineeringCenter/HEC-FDA/assets/47757848/1fbaa01f-7c2a-43bf-a2c5-830ebaa023fb)

codymccoy:
@rnugent3 Based on the pics above it looks like you want me to change the checkbox text depending on if it is selected or not. Is that correct?
if unselected: "Include Non-Breach Risk"
if selected: "Breach AND Non-Breach Conditions Uniquely Modeled"

rnugent3:
Sorry for the confusing text. Let's have a static label "Non-Failure Scenario Modeled"

rnugent3:
And please change "breach" and "no breach" to "Failure" and "Non-Failure"

codymccoy:
![image](https://github.com/HydrologicEngineeringCenter/HEC-FDA/assets/47757848/c1eebb45-a43b-4870-af3a-8b4773cc5043)

rnugent3:
Looks great. 

The validation on the requirement of a levee ... in case you haven't caught it already ... we should check that at least one impact area has a levee

codymccoy:
I am refreshing on how validation works in this editor. Looks like we no longer validate before saving. We allow the users to save in an error state.
![image](https://github.com/HydrologicEngineeringCenter/HEC-FDA/assets/47757848/29973403-d4fc-402e-8aef-86b3ac7a5de2)

I have an icon on the tabs that also have a tooltip if in error:
[...]

PR #967, opened 2023-09-20
956 UI updates

codymccoy:
this should complete issue #963


rnugent3:
@codymccoy I have one tiny request. when the checkbox is not marked, could the stage-damage combo box label be "Stage-Damage" while if the checkbox is marked, let the combo box label be "Failure Stage-Damage"

rnugent3:
Otherwise, the UI edits look good to me. I've not yet run because we're not yet hooked up. @Brennan1994 would you mind taking a look at the code changes?

codymccoy:
Sure, i can make that change.

codymccoy:
I just pushed the requested label change. I also moved the red stars down a bit on the new stage damage row.

Brennan1994 (approved):

PR #998, opened 2023-10-13
995 paired data search broken

rnugent3:


rnugent3:
I think that this work is the last of what remains before we're ready to send to the PCX.

HenryGeorgist (approved):
looks good to me

PR #1001, opened 2023-11-03
Address difficult to diagnose user errors when with importing SI 

Brennan1994:
These updates address issues raised by @rtle76 when setting up a project.

These changes do 3 things
- We limit the file selector to HDF and TIF in the Terrain importer. 
- We now support Tif in the Terrain importer
- We now always reproject to the terrain projection automatically rather than relying on the user entered projection as long as there is a terrain available.


Brennan1994:
addressing issues I've observed from Dean and @rtle76 as part of informal software review and Dean's formal review.

PR #1013, opened 2023-11-20
Bug/previewcompute

rnugent3:
Two bug fixes here: one in which I did not pass curve meta data through the graphical UPD curve sampling, and the other where we had some logic issues in dealing with damage categories in the scenario editor.


rnugent3:
Brennan, please hold off on reviewing, I forgot to do my cleanup.

rnugent3:
It turns out that the non-fail stage damage functions are being handled completely different from the default stage damage functions. For example, they exist as different types. 

I have added in validation to check whether there is a non fail, but we need to refactor these classes so that the two sets of stage damage functions are being handled similarly.

Brennan1994 (approved):

PR #1031, opened 2023-12-11
Bugs/stage damage

rnugent3:


HEC.FDA.Model/stageDamage/ImpactAreaStageDamage.cs
```diff
@@ -38,8 +38,8 @@ public class ImpactAreaStageDamage : ValidationErrorLogger, IContainValidationGr
         private double _MinStageForArea;
         private double _MaxStageForArea;
 
-        private int _NumExtrapolatedStagesToCompute;
-        private int _NumInterpolatedStagesToCompute;
+        private int _MoreInterpolationPoints;
```
Brennan1994, line 41: This isn't critical, but  I'd like a comment explaining what these are, and what they're used for. It's pretty tough to follow without.

Brennan1994 (approved):

PR #1041, opened 2023-12-21
Projection probs

Brennan1994:


Brennan1994:
Very likely the projection issue we've been seeing stem from a RASMapper bug that was struggling to get the files from the terrain. I suspect we'll have better luck now, but explicit testing of the projections need to be done. Added additional validation on the terrain file. Makes sure we have component files. 


image

PR #1049, opened 2024-01-05
added stack trace to stage damage exception catch

rnugent3:
teeny tiny change


rnugent3:
I broke something in view results...please hold off review

rnugent3:
This PR also includes the code required to report progress on stage damage functions from the model, and a couple format changes

rnugent3:
@Brennan1994 this is ready for your review

rnugent3:
still broken

rnugent3:
When we do not have any structures in a given impact area, we do not produce stage-damage functions. However, it is perfectly reasonable that we need stage damage functions that have zero damage for comparison with stage damage that has positive damage if an impact area is being bought out

rnugent3:
I need to add validation to make sure that an impact area scenario either has minimally stage-damage functions or a levee

Brennan1994 (approved):

PR #1060, opened 2024-02-03
Better histogram bin specification for stage-damage

rnugent3:
I used to look at the scale of the maximum value in the first sample to inform bin width. 

Now, we just start with 500 bins. More will be added if necessary.


Brennan1994:
I think this may be significant to our memory problem. Test not passing, and I'm getting exceptions when testing with real data though. I'm going to give it some work tomorrow too.

Brennan1994 (approved):

PR #1065, opened 2024-02-27
Bug/bin

rnugent3:
restrict stage damage bin width between 1 and 100.


rnugent3:
@Brennan1994 ready for your review

PR #1077, opened 2024-03-23
Update SQL version and Dynamic Histograms

rnugent3:


rnugent3:
most changes are a name change, was excited to level that up, all relevant changes are in the histogram class

rnugent3:
this is probably going to give us a nice speed boost

PR #1089, opened 2024-04-27
Fix #1088

### LINKED ISSUE #1088: Error When Reopening Previously Saved Study
This error is happening again/still: 

image

It doesn't seem to be happening with studies that are 1.4.3 exports to 2.0, and does seem to be happening when it's a full 2.0 model.

Brennan1994:


rnugent3:
ready for review?

rnugent3:
on it

rnugent3 (approved):
Looks good. 

My curiosity makes me wonder how we could have been able to open the file and then the same day have XML issues. 

I might be making things up, but it seems like there could have been a file writing problem, which still might be there, and if so and the problem shows up again, then we won't try to import the data?

rnugent3:
I guess my question is - did we address the root cause or do we have a fix for a symptom?

Brennan1994:
> I guess my question is - did we address the root cause or do we have a fix for a symptom?

Got rid of that todo*

We did fix the system. the fromXML for hydraulic profiles requires a profile name. I wasn't saving gridded data with a profile name, because it doesn't need one, but if you save an attribute to an xml element as null, XElement completely ignores that attribute all together.  

Now I check that the attribute exists before I get it's value. Admittedly, we should be following that pattern more often to avoid these problems.

PR #1098, opened 2024-05-15
Griddataimp

Brennan1994:


HEC.FDA.ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -203,26 +223,30 @@ private void SaveNew()
...
             }
 
             int id = GetElementID<HydraulicElement>();
-            HydraulicElement elementToSave = new(Name, Description, pathProbs, HydraulicDataSource.WSEGrid, id);
+            HydraulicElement elementToSave = new(Name, Description, hydraulicProfiles, HydraulicDataSource.WSEGrid, id);
             base.Save(elementToSave);
         }
 
-        private string getFilePathFromChildElement(WaterSurfaceElevationRowItemVM row)
+        /// <summary>
```
Brennan1994, line 241: This method was a mess. It should get the file path of the tif from child element root. None of the naming conventions conveyed that, and the method for actually gettting it was convoluted.

HEC.FDA.ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -203,26 +223,30 @@ private void SaveNew()
...
             Directory.CreateDirectory(destinationDirectory);
-            List<HydraulicProfile> pathProbs = new();
+            List<HydraulicProfile> hydraulicProfiles = new();
             foreach (WaterSurfaceElevationRowItemVM row in ListOfRows)
             {
-                string fileNameFromChildElementDir = getFilePathFromChildElement(row);
-                pathProbs.Add(
-                    new HydraulicProfile(row.Probability, fileNameFromChildElementDir, null));
-                StudyFilesManager.CopyDirectory(row.Path, row.Name, destinationDirectory);
+                string fileNameFromChildElementDir = GetFilePathFromChildElementRoot(row);
```
Brennan1994, line 229: This change was for consistency largely. We consider row.Path in every other importer to be the absolute path to the file that contains the data. Prior to this refactor, in gridded data it actually reffered to the directory, not the file. Because of the refactor I now pull out the source directory from row.path before passing it into the StudyFilesManager.Copy method.

HEC.FDA.ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -174,25 +194,25 @@ public override void Save()
...
         private void RenameDirectoryInTheStudy()
         {
-            if (!Name.Equals(OriginalElement.Name))
-            {
-                string sourceFilePath = Connection.Instance.HydraulicsDirectory + "\\" + OriginalElement.Name;
-                string destinationFilePath = Connection.Instance.HydraulicsDirectory + "\\" + Name;
-                Directory.Move(sourceFilePath, destinationFilePath);
-            }
+            if (Name.Equals(OriginalElement.Name)) { return; }
+            string sourceFilePath = Connection.Instance.HydraulicsDirectory + "\\" + OriginalElement.Name;
```
Brennan1994, line 203: Stylistic change. for voids, I generally prefer to early exit with if statements, rather than put all the logic in the if statement. keeps ups one set of brackets deep.

HEC.FDA.ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -139,11 +158,12 @@ public void FolderSelected(string fullpath)
...
                     }
                 }
-                double prob = 0;
                 foreach (string dir in validDirectories)
                 {
-                    prob += .1;
-                    AddRow(Path.GetFileName(dir), Path.GetFullPath(dir), prob);
+                    string name = Path.GetFileName(dir);
+                    string[] tifFiles = Directory.GetFiles(dir, "*.tif");
+                    string tif = tifFiles[0]; //there should only be one, but if there are more, we'll just take the first one
```
Brennan1994, line 165: This enforces the Path property of a row actually being the full path to the tif. That's consistent with the other importers.

HEC.FDA.ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -113,6 +121,17 @@ private FdaValidationResult ValidateTIFFiles(List<string> tifFiles, string direc
...
         #endregion
 
+        /// <summary>
+        /// Creates a VM row from the HydraulicProfile. This is used when editing an existing element.
+        /// </summary>
+        private static WaterSurfaceElevationRowItemVM CreateRowFromHydraulicProfile(IHydraulicProfile profile, string elementName)
+        {
+            string name = Path.GetDirectoryName(profile.FileName); //name is what's shown to the UI. For Gridded data, we want the directory name, not the file name. 
+            string path = Path.Combine(Connection.Instance.HydraulicsDirectory, elementName, profile.FileName); //path is the full path to the file from the element root.
+            double prob = profile.Probability;
```
Brennan1994, line 131: We actually save hydraulic profiles, not these rows to database, which is how it should be. So I'm just creating these rows from the saved data objects here.

HEC.FDA.ViewModel/Hydraulics/GriddedData/GriddedImporterVM.cs
```diff
@@ -42,11 +44,15 @@ public GriddedImporterVM(EditorActionManager actionManager) : base(actionManager
...
         /// Constructor used when editing an existing child node.
         /// </summary>
-        /// <param name="elem"></param>
-        /// <param name="actionManager"></param>
         public GriddedImporterVM(HydraulicElement elem, EditorActionManager actionManager) : base(elem, actionManager)
         {
-            SelectedPath = Connection.Instance.HydraulicsDirectory + "\\" + elem.Name;
+            IEnumerable<IHydraulicProfile> profiles = elem.DataSet.HydraulicProfiles.OrderBy(x => x.Probability);
+            ListOfRows.Clear();
+            foreach (IHydraulicProfile profile in profiles)
```
Brennan1994, line 51: This was the source of our bug. I don't know why I thought the prior would work. That was definitely my doing. now we're actually loading the profiles from the element.

Brennan1994:
Merging without review for expedience.

PR #1100, opened 2024-05-28
recreate distributions with histograms

Brennan1994:
recreated with the help of github copilot


Brennan1994:
This is failing on lognormal. I can nail the quantiles, but not pull out a mean and standard deviation. We should discuss.

rnugent3 (changes requested):
remove lognormal, add documentation to triangular assertion that median is mode for test case

rnugent3:
we have many tests that demonstrate that log normal works correctly, and we do not support log normal in the economics so our log normal class is less relevant for this purpose

rnugent3 (approved):

PR #1103, opened 2024-06-06
Dot net8

Brennan1994:


Brennan1994:
This was some work. The big issue is the getting rid of binary formatter. That was a really special hack we were using for a long time to deep copy objects. Relying on JSON now, we're only copying public fields and props. it's working for occupancy types, but I'm afraid it will fail when we try to import agg stage damage curves. Need a point to test data, or a hand testing.

Brennan1994:
confirmed Agg stage damage importer still seems to work. I think this solves the import problem, along with some of the geospatial issues now. 
@rnugent3 , Ready for review.

HEC.FDA.ViewModel/Utilities/ImportFromFDA1Helper.cs
```diff
@@ -858,7 +858,7 @@ private static UncertainPairedData CreateEmptyFunction()
 
         private static bool IsEmptyFunction(SingleDamageFunction function)
         {
-            return function.NumOrdinates == 0;
+            return false;
```
Brennan1994, line 861: This looks wrong.

HEC.FDA.Model/hydraulics/HydraulicDataset.cs
```diff
@@ -65,21 +65,34 @@ public bool LoadFromXML(XElement xElement)
...
+            PointFeatureLayer indexPoint = new("ThisNameIsNotUsedForAnythingHere", pointShapefileFilePath);
             if (!indexPoint.SourceFileExists)
             {
                 return null;
             }
-            PointMs indexPoints = new PointMs(indexPoint.Points().Select(p => p.PointM()));
+            PointMs indexPoints = new(indexPoint.Points().Select(p => p.PointM()));
+            
+            //reproject the point to the study projection. For GetWSE, all the reprojection is done in the Inventory class. We aren't using an inventory here, so we need to do it manually.
+            Projection pointProj = RASHelper.GetVectorProjection(indexPoint);
```
Brennan1994, line 83: This is a substantial change from what we did before. Seems correct to me, but could be the change?

HEC.FDA.Model/hydraulics/HydraulicDataset.cs
```diff
@@ -65,21 +65,34 @@ public bool LoadFromXML(XElement xElement)
...
+            if (!pointProj.IsEqual(studyProjection))
+            {
+                indexPoints.Project(pointProj,studyProjection);
```
Brennan1994, line 86: another change to check.

HEC.FDA.Model/hydraulics/HydraulicDataset.cs
Brennan1994: No. Red herring. I think this is fine.

HEC.FDA.Model/hydraulics/HydraulicDataset.cs
Brennan1994: another red herring. probably fine.

HEC.FDA.ViewModel/Utilities/ImportFromFDA1Helper.cs
Brennan1994: Fixed this.

PR #1127, opened 2024-08-15
Adding logic to check that types match expected when loading SI from …

Brennan1994:
…shapefile.


rnugent3:
Does this change mean that we can import long formatted data successfully, or that the software will know to reject long formatted data?

Brennan1994:
Right now this just rejects it on SI import. I'd like to just accept it and cast it, but it's trickier than you might expect. I'm working on it.

Brennan1994:
we now reject long formatted data. Merging this in for now, as it's a better system then before. Ideally we'd accept any number type for number values.

PR #1131, opened 2024-08-21
Fixes #1129

### LINKED ISSUE #1129: CSVR Normal Distribution Import from TXT File Error
CSVR with a normal distribution of mean 114 and CV of 71.5 imported from TXT

image

Should result in normal distribution with a mean of 114 and st dev of 81.51

Does not: 

image

rnugent3:
Related Work Items: #1129


Brennan1994:
What's going on here? I see you've just deleted a method that intended to convert. Did it not need converting all along?

rnugent3:
@Brennan1994 please take a look at the issue? I think the answers to your questions are there.

Brennan1994 (approved):

PR #1141, opened 2024-09-04
fix validation of deterministic CSVRs

### LINKED ISSUE #1140: Deterministic CSVR Doesn't work

Brennan1994:
resolves #1140


Brennan1994:
I've walked this through a compute and seen success on my end now. If you can run it on the original offender, that'd confirm we're out of the woods for me. 

Shoddy validation being the culprit.

rnugent3 (approved):
I can confirm that we're calculating damage for the offending study data. I also have to share that we're now calculating damage for the offending study data based on the main build, too, so the problem with the offending study data was not this issue.

PR #1143, opened 2024-09-18
Remove fit to flows

Brennan1994:


Brennan1994:
includes backward compatability logic to migrate forward projects which did include fit to flows data. 

This does not save any of that data if they had data in there, and does not continue to write it. I think this is fine for an unreleased software update.

PR #1144, opened 2024-09-19
Object Copier was failing for rating function. Added specific impleme…

Brennan1994:
fix#1128  The issue seems to have come from the object copier failing to deep copy rating functions. This deserves further investigation for the other functions, but this solution solves the problem at hand.


rnugent3 (approved):
works nicely

PR #1149, opened 2024-09-27
1104 create a new control to view software version in the UI

Brennan1994:
Ok. A lot happened here. 

Importantly: 

1. Version now shows in the title bar of the main window. Used to say FDA 2.0, now says the actual version. I dig this over having an about me or "about" context menu somewhere. Right out front. super obvious. 

2. The version you compute with is now saved to a scenario result. Satisfying Britt's suggestion. 

3. Workflows updated to automatically version the software. 

4. Spring cleaning on csproj files and unused projects. From now on, anything we need a console app for rather than a unit test, we use Scratch.csproj. We can talk about how that works next time we're in person.


Brennan1994:
Turns out I don't like my solution as much as i thought. Microsoft doesn't include beta/alpha tags in the file version, and throws a bunch of alphanumerics in the product verison. Needs more love. We're close. Maybe over the weekend.

Brennan1994:
New idea. We use the first 3 version numbers. If there's a fourth (like a build counter, that I've implemented from CI builds) it'll include a Beta in the title bar. 

If it's a release, we won't add a build number. the 2.0 release is 2.0.0. and it will show that way in the window and in results.  I think this is pretty slick.

rnugent3:
@Brennan1994 where among the results do you report the version with which those results were produced?

Brennan1994:
image

rnugent3 (approved):

BrittCorley:
> image
> 

We'll have to make sure users know that they will need to manually add the version to any result tables they copy and paste to make their appendix tables.

rnugent3:
@BrittCorley is that new to the FDA --> Economic Appendix workflow?

rnugent3:
In other words, did you copy and paste result tables from 1.4.3, result tables which included the version number?

rnugent3:
maybe a little too pointed lol 

and I get your point, where would you suggest we add that advice to the user documentation?

BrittCorley:
> maybe a little too pointed lol 
> 
> and I get your point, where would you suggest we add that advice to the user documentation?

@rnugent3 lol not you coming at me like that damn

I was thinking we could add a section to the user manual about interpreting results and reporting data that shows them what they need to report, where to find what they're looking for, and how to translate into 5th grade language. We talked about it once upon a time when I was complaining about the project performance being hard to communicate. One of the many things on the list of stuff Britt wants to do.

PR #1152, opened 2024-09-30
We need more compute points at the bottom for weird situations

rnugent3:
I had programmed the aggregated stage-damage compute to use less compute points between the most frequent event in the hydraulics data set and the .999 stage, but it turns out we need more compute points because we can miss important nonlinearity beyond the most frequent event in the hydraulics data set when there is positive damage at that event. 

Before the fix, there is a 4-ft gap between the 0.5 AEP stage and the next more frequent stage. 
image

After the fix, the difference is about 0.5 ft. 
image


rnugent3:
We got such a big range of stages at the bottom of the stage-damage function for this study because there is a 0 discharge stage in the stage-discharge function.
image

rnugent3:
Hold off on reviewing. I need to think through a more flexible approach. We'll be adding many compute points for many situations where they do nothing but add computational expense.

rnugent3:
@Brennan1994 this is ready for your review

HEC.FDA.Model/stageDamage/ImpactAreaStageDamage.cs
```diff
@@ -80,37 +81,50 @@ public ImpactAreaStageDamage(int impactAreaID, Inventory inventory, HydraulicDat
...
-            //require at least two coordinates to interpolate and extrapolate 
-            if (coordinateQuantity < 4)
-            {
-                coordinateQuantity = 4;
-            }
-            _MoreInterpolationPoints = coordinateQuantity * 4;
-            _LessInterpolationPoints = coordinateQuantity;
-
+            //0.25ft target more than preserves the level of information content 
+            double feetPerCoordinate = 0.25;
```
Brennan1994, line 100: pull this up as class level constant

Brennan1994 (commented):
pull constants out and leave a comment on them to describe why they were chosen.

Brennan1994 (approved):

PR #1154, opened 2024-10-10
Remove db manager

jackschonherr:


HEC.FDA.ViewModel/Saving/SavingBase.cs
```diff
@@ -289,15 +293,15 @@ public void DeleteRowWithKey( int key, string keyColName)
...
+        }
 
         #endregion
 
         public int GetNextAvailableId()
         {
             //make sure the table exists
             OpenConnection();
-            DatabaseManager.DataTableView tbl = Connection.Instance.GetTable(_TableName);
+            object tbl = Connection.Instance.GetTable(_TableName);
```
Brennan1994, line 304: This is weird. If we're just confirming a table exists, can we do that with a bool instead of returning an object?

HEC.FDA.ViewModel/Storage/Connection.cs
```diff
@@ -275,7 +320,7 @@ public string[] TableNames()
         {
             return _SqliteReader.GetTableNames();
         }
-        public DatabaseManager.DataTableView GetTable(string TableName)
+        public object GetTable(string TableName)
```
Brennan1994, line 323: From what i can tell, this method is only ever used to check whether a table exists. Wrap this method in another method called "TableExists(string tableName)" and call this internally. It'll make it a lot clearer how it's used.

HEC.FDA.ViewModel/Saving/SavingBase.cs
```diff
@@ -274,13 +278,13 @@ public void DeleteRowWithCompoundKey(string tableName, int[] primaryKeys, string
             command.ExecuteNonQuery();
         }
 
-        public void DeleteRowWithKey( int key, string keyColName)
+        public void DeleteRowWithKey(int key, string keyColName)
```
Brennan1994, line 281: Not seeing references in this guy. Confirm that. might be able to delete.

Brennan1994 (changes requested):

Brennan1994:
There's quite a few functions in Saving Base and connection that appear to have no references. Verify that and remove them if that's true.  SavingBase.DeleteRowWithKey, SavingBase.DeleteRowWithCompoundKey, SavingBase.UpdateTableRowWithCompoundKey all appear that way

HEC.FDA.ViewModel/Saving/SavingBase.cs
jackschonherr: can confirm, removed references

HEC.FDA.ViewModel/Storage/Connection.cs
jackschonherr: done, makes it clearer

HEC.FDA.ViewModel/Saving/SavingBase.cs
jackschonherr: yes, replaced with a bool wrapper function

Brennan1994 (approved):

rnugent3:
@Brennan1994 @jackschonherr - thank you so much for all the hard work!

PR #1164, opened 2024-11-05
reset histograms to min of zero on the fly

rnugent3:
I created a fix that shifted a histogram if we obtained a zero-valued result after construction of a histogram to allow the **Histogram** Min to be 0, not just the sample min. The fix requires us to shift the histogram up to add the required number of bins to reach zero, and put the leftward most edge of the histogram right at 0. This means that pre-shift bins and post-shift bins will partially overlap, and those partials need to be allocated out correctly based upon the fraction of the observations in a pre-shift bin that overlaps with the post-shift bin, where the fraction is determined with the assumption of a uniformly distributed bin. I think the majority of the histogram math works that way. 

This fix is only reached if the min is 0. If the min goes below 0, so will the histogram. 

This fix produces a really nice result. Prior to the fix, the histograms gave us really wonky left tails. See for example: 
image
After the fix, we have a logical left tail:
image

A unit test is included to demonstrate that the added logic works as expected.


rnugent3:
Security systems are preventing me from running a unit test locally so I pushed it here

Brennan1994 (approved):

PR #1173, opened 2024-11-22
Output table updates

Brennan1994:


Brennan1994:
The code diff here is miserable. There was a lot just shuffled around that's showing up as new code. This is probably easiest smell tested in the UI, then if something seems funny dive into source. Should have everything we talked about incorporated now.

rnugent3 (changes requested):
I recommend the following changes to the alternatives summary results: 
- Include acronym in title (EqAD)
- Consistent acronym use in column headers
image

I recommend the following changes to alternative comparison report summary results tables: 
- include sufficient arguments of WOP EAD and EqAD to get alternative, impact area specific metrics 
- Include "Reduced" in all table titles 
- Match EqAD table title and column headers with alternative summary results 
image

Addition of other asset categories shows up nicely. If for a given scenario an additional asset category is used, we'll display metrics for that asset category for all damage categories and impact areas. The asset category doesn't show up for other scenario reporting. I think this works nicely. 
image

Brennan1994:
I believe everything is addressed. Give it another look. Thanks for catching those mistakes,

rnugent3 (approved):
Tables look great. Thank you, @Brennan1994.

PR #1185, opened 2024-12-02
Update RecommendationExpanderControl.xaml

Brennan1994:


rnugent3 (approved):
I have confirmed that the exception no longer takes place.

PR #1190, opened 2024-12-10
closes #1187

### LINKED ISSUE #1187: Dummy occupancy type inventory assignment saves incorrectly
Bug report from the field: 

`I am encountering a bug in 11/18/24 version of FDA 2.0 when using manually entered occupancy types. When I import a structure inventory the Default Occupancy Type Assignment is saving incorrectly. 

They’re importing correctly, as seen in the first screenshot but when I click Finish the inventory is imported but all the occupancy type assignments change to MULTI-PIER, which is the first occupancy type I added to when creating the occupancy type set. This issue has occurred for new and existing inventories, for all created Occupancy Types, and in new projects. Imported Occupancy Types save correctly, is there a way to export my Occupancy Type set to a text file?

I’ll show some of my trouble shooting using a dummy Occupancy Type set.
 

I created a dummy Occupancy Type set, entered Value 1 and Value 2, clicked save on the occupancy type set, entered Value 3, then saved and closed. I edited my inventory to manual assign occupancy types in the order seen below. 
 

When I click finish to import the inventory, then re-open the inventory to edit the screenshot below shows what I see. Value 2’s assignments are replaced by Value 1. For some reason Value 3 is not replaced which was unexpected since this did not happen for my other occupancy type sets. Values 1 and 3 can be moved around and will save correctly, value 2 is always overwritten.
 

Let me know if I can provide any further clarification. I will upload this model to the Study Drop-off folder, its titled “LC205 FDA v2”.`

Brennan1994:


Brennan1994:
I believe the behavior described before was due to bad occtype ID's being assigned. New code increments the last ID from the group by 1, which i I think was the intent of the original author.

rnugent3 (changes requested):
Upon creating a new occupancy type within the UI, we now hit an unhandled exception

image

`
image
`

Brennan1994:
@rnugent3 that change should have fixed the issue. Along with it I removed the tracking of group id from occtypes. Occtypes don't need a group ID because they're children of a group, with an ID already. It was redundant, and leading to situations where they dissagreed.

rnugent3 (approved):
Behaves as expected. Thank you, @Brennan1994

PR #1204, opened 2025-01-09
carry price index through when trimming to impact area.

Brennan1994:


HEC.FDA.ModelTest/unittests/StageDamageShould.cs
```diff
@@ -319,6 +319,25 @@ public void ProduceNonNullResults()
 
         }
 
+        [Theory]
+        [InlineData(1,2600)]
+        [InlineData(2,5200)]
```
rnugent3, line 324: great test

rnugent3 (approved):
The price index is showing up nicely in the structure stage damage details, as expected

image

PR #1206, opened 2025-01-13
Fixes issue swapping projections

Brennan1994:
fixes an issue where an exception was thrown when setting a different projection after a projection was already set for the study. New behavior is to archive the prior in a subfolder, delete it from the projection folder, and copy the new file in.


rnugent3 (approved):
I've confirmed that I can change a projection file without hitting an exception.

PR #1207, opened 2025-01-14
Enhancement/freq curve visualization

Brennan1994:
This PR 

- Now only calculates tabulated data for analytical frequency curves upon opening an editor (rather than on study open) which reduces startup time
- Updates confidence bounds on graphical frequency upon editor open.
- Various code cleanup


rnugent3 (approved):
Changes look good to me. Further, I have confirmed that opening a study with 1000 graphical frequency curve stalls the software, and that I could open the study with the changes on this branch.

PR #1209, opened 2025-01-22
fixes issue with radio buttons not loading on initial open, as well a…

Brennan1994:
This PR does the following
- updates a view dependency to latest version
- changes the github workflow file to no longer authenticate for the RAS nuget dependency (it's now public read-only)
- fixes a bug where having multiple frequency curves open would sometimes show neither flow or stage selected for graphical 
- fixes a bug where graphical frequency curves (stage) would load with no radio button checked.


rnugent3 (approved):
Code changes look good and software behaves as intended

PR #1216, opened 2025-02-24
Stage damage data issues

Brennan1994:


Brennan1994:
This PR needs to do the following before merge
- Improve readability of shapefile loading
- Screen out data from unselected menu's in the structure inventory editor (if using terrain for elevs, don't send the FFE to model)
- Allow users to clear selections in the SI combo boxes.

rnugent3 (approved):
Confirmed that software behaves as intended

PR #1217, opened 2025-02-24
Bug/fix alternative compute

Brennan1994:


Brennan1994:
failure related to nuget package sources.

Brennan1994:
This PR addresses an issue with computing and getting message back from the alternative compute. The compute is wicked fast, so the progress bar should go relatively unnoticed, but this adopts some more of RAS's technology to include their progress reporter.  This adds a dependency on another RAS Repo which does require credentials, so be aware of that when you test. 

Should address the attached issues.

rnugent3 (approved):
Confirmed that this PR fixes the validation

PR #1221, opened 2025-03-09
Geospatial update

Brennan1994:
This update moves us largely off RasMapper for geospatial operations.  We do still use PointM as our coordinate for structures on construction. That's important because we need to work in old ras space (RasMapper) to use RAS6.x terrains and results. So everywhere we can, we're now sticking to the actively developed RAS2025 codebase and ways of doing things, and only reverting for operations we can't support.

### DIFF
--- a/HEC.FDA.Model/Spatial/Extensions/Utility.cs
+++ b/HEC.FDA.Model/Spatial/Extensions/Utility.cs
@@ -9,7 +9,9 @@ public static class Utility
     public static T TryGetValueAs<T>(this TableRow row, string columnName, T fallbackValue = default) {
 		try
 		{
-			return row.ValueAs(columnName, fallbackValue);
+			if(string.IsNullOrWhiteSpace(columnName))
+                return fallbackValue;
+            return row.ValueAs(columnName, fallbackValue);
         }
 		catch (System.Exception)
 		{
--- a/HEC.FDA.Model/Spatial/Extensions/Utility.cs
+++ b/HEC.FDA.Model/Spatial/Extensions/Utility.cs
@@ -1,0 +1,19 @@
+﻿using Utility.Memory;
+
+namespace HEC.FDA.Model.Spatial.Extensions;
+public static class Utility
+{
+    /// <summary>
+    /// Tries to get the value of the specified column as the specified type. If the type is wrong, or column doesn't exist, returns fallback value
+    /// </summary>
+    public static T TryGetValueAs<T>(this TableRow row, string columnName, T fallbackValue = default) {
+		try
+		{
+			return row.ValueAs(columnName, fallbackValue);
+        }
+		catch (System.Exception)
+		{
+			return fallbackValue;
+		}
+    }
+}
--- a/HEC.FDA.Model/Spatial/StructureFactory.cs
+++ b/HEC.FDA.Model/Spatial/StructureFactory.cs
@@ -14,7 +14,7 @@ using Utility.Memory;
 using HEC.FDA.Model.Spatial.Extensions;
 
 namespace HEC.FDA.Model.Spatial;
-public class ShapefileLoader
+public class StructureFactory
 {
     private const string DEFAULT_MISSING_STRING_VALUE = "EMPTY";
     private const int DEFAULT_MISSING_NUMBER_VALUE = IntegerGlobalConstants.DEFAULT_MISSING_VALUE;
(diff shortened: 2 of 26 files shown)


HEC.FDA.Model/Spatial/Extensions/Utility.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace HEC.FDA.Model.Spatial.Extensions;
+public static class Utility
+{
+    /// <summary>
+    /// Tries to get the value of the specified column as the specified type. If the type is wrong, or column doesn't exist, returns fallback value
+    /// </summary>
+    public static T TryGetValueAs<T>(this TableRow row, string columnName, T fallbackValue = default) {
+		try
+		{
+			if(string.IsNullOrWhiteSpace(columnName))
```
Brennan1994, line 12: ValueAs throws if we ask for a column that doesn't exist. This stops that throw when we check for mapped attributes that weren't set (come in as empty strings for column names). Throwing exceptions is super expensive though, so we check full null or whitespace and gracefully "fail" first.  The catch shouldn't ever hit without something exceptional happening, like someone modifying the column names of the shapefile we're using after import.

HEC.FDA.Model/Spatial/StructureFactory.cs
```diff
@@ -0,0 +1,150 @@
...
+using Geospatial.Features;
+using Geospatial.IO;
+using Geospatial.Vectors;
+using Utility.Logging;
+using Geospatial.GDALAssist.Vectors;
+using Utility.Memory;
+using HEC.FDA.Model.Spatial.Extensions;
+
+namespace HEC.FDA.Model.Spatial;
+public class StructureFactory
```
Brennan1994, line 17: This is a stupid visualization of a class rename. This used to be ShapefileReader, which wasn't descriptive enough.

HEC.FDA.Model/structures/Inventory.cs
```diff
@@ -40,8 +41,16 @@ public Inventory(string pointShapefilePath, string impactAreaShapefilePath, Stru
         {
             OccTypes = occTypes;
             PriceIndex = priceIndex;
-            Projection studyProjection = Projection.FromFile(projectionFilePath);//Projection.FromFile returns Null if the path is bad. We'll check for null before we reproject. 
-            Structures = ShapefileLoader.LoadStructuresFromSourceFiles(pointShapefilePath, map, null, false, impactAreaShapefilePath, studyProjection, OccTypes);
+            Projection studyProjection = Projection.FromFile(projectionFilePath);//Projection.FromFile returns Null if the path is bad. We'll check for null before we reproject.
+            OperationResult operationResult = StructureFactory.LoadStructuresFromSourceFiles(pointShapefilePath, map, null, false, impactAreaShapefilePath, studyProjection, OccTypes, out List<Structure> structures);
```
Brennan1994, line 45: Setting up for better messaging by returning an operation result which has messages in it to explain failures.

HEC.FDA.Model/structures/Structure.cs
```diff
@@ -39,8 +42,12 @@ public class Structure: Validation
         #endregion
 
         #region Constructors 
-        public Structure(string fid, PointM point, double firstFloorElevation, double val_struct, string st_damcat, string occtype, int impactAreaID, double val_cont =0, double val_vehic = 0, double val_other = 0, string cbfips = "unassigned", double beginDamage = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, double groundElevation = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, double foundationHeight = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, int year = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, int numStructures = 1, string notes = "", string description = "")
-
+        /// <summary>
+        /// Maintained to support point M.
+        /// </summary>
+        public Structure(string fid, PointM point, double firstFloorElevation, double val_struct, string st_damcat, string occtype, int impactAreaID, double val_cont =0, double val_vehic = 0, double val_other = 0, 
+            string cbfips = "unassigned", double beginDamage = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, double groundElevation = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE,
```
Brennan1994, line 49: reformat.

HEC.FDA.Model/structures/Structure.cs
```diff
@@ -62,6 +69,33 @@ public Structure(string fid, PointM point, double firstFloorElevation, double va
...
             Description = description;
         }
+        /// <summary>
+        /// Preferred Constructor for now. Eventually will store Geospatial.Vectors.Point instead of PointM.
+        /// </summary>
+        public Structure(string fid, Geospatial.Vectors.Point point, double firstFloorElevation, double val_struct, string st_damcat, string occtype, int impactAreaID, double val_cont = 0, double val_vehic = 0, double val_other = 0, string cbfips = "unassigned", double beginDamage = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, double groundElevation = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, double foundationHeight = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, int year = utilities.IntegerGlobalConstants.DEFAULT_MISSING_VALUE, int numStructures = 1, string notes = "", string description = "")
+
+        {
+            Fid = fid;
+            PointM pt = Converter.ConvertPtM(point);
```
Brennan1994, line 79: Here we're bouncing back to PointMs. We're married to these for sampling from RAS6 things.  I think our best path forward is to always stay in 2025 land, because we have better access to fix problems directly through me, and through relationships with the team on that project, rather than 6.x, which isn't being maintained to the same degree in C#.

HEC.FDA.ViewModel/HEC.FDA.ViewModel.csproj
```diff
@@ -13,7 +13,7 @@
 	</ItemGroup>
 	<ItemGroup>
 		<PackageReference Include="CommunityToolkit.Mvvm" Version="8.4.0" />
-		<PackageReference Include="HEC.RAS.Visual.Observables" Version="0.1.0.1421-dev" />		
+		<PackageReference Include="HEC.RAS.Visual.Observables" Version="0.1.0.1554-dev" />
```
Brennan1994, line 16: Jumping to latest RAS.

nuget.config
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
 	<packageSources>
-		<clear/>
```
Brennan1994, line 4: Do you get prompted for creds every time you open  a project? I suspect this clear was part of the problem. It's unclear though. I'd love to track this down.

rnugent3 (approved):
Code changes look good. Changes have been tested by the field with success.

PR #1222, opened 2025-03-10
Fix Copy Paste 

Brennan1994:
Address an issue where copying from a datagrid in FDA would not paste into Excel. I have not been able to track down when, or why this happened. "Copy" was handled by the System.Windows Library, not us. My best guess is that the behavior changed between Net 6.0 -< Net 9.0, but I have not yet found the commit where we changed, or found discussion on the webs. 

Nevertheless, this is a pretty clean fix as far as hacks go. Windows copies data in multiple formats when you call copy. I grab the data in "text" format, purge the clipboard, and set the text data back to the clipboard every copy, so the data we want for excel is the only data in there.

seems to work.

### DIFF
--- a/HEC.FDA.View/TableWithPlot/FdaDataGrid.cs
+++ b/HEC.FDA.View/TableWithPlot/FdaDataGrid.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using System.Windows;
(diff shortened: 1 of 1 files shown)


Brennan1994:
Bypassed normal approval. value of delivering a better functioning beta outweighted the risks of this 3 line change.

rnugent3:
I have confirmed that copy and paste works right right click, copy. Copy and paste still fails when copying through CTRL +C.

PR #1223, opened 2025-03-11
1214 uncertainty around graphical stage frequency appears to be incorrect

rnugent3:
I have redesigned the computational algorithm for generating uncertainty about graphical frequency curves and have added a unit test for the edge case (provided by the Algiers PDT) that illuminated the need for the change. HEC-FDA Version 2.0 is now producing nearly exactly the same uncertainty as HEC-FDA Version 1.4.3 for this edge case.

### DIFF
--- a/HEC.FDA.Model/paireddata/PairedData.cs
+++ b/HEC.FDA.Model/paireddata/PairedData.cs
@@ -366,7 +366,31 @@ namespace HEC.FDA.Model.paireddata
             }
             Yvals = update;
         }
-        public void ForceStrictMonotonicity(double max = double.MaxValue, double min = double.MinValue)
+        public void ForceStrictMonotonicityTopDown(double max = double.MaxValue, double min = double.MinValue)
+        {
+            double epsilon = 0.005;
+
+            double[] update = new double[Yvals.Length];
+            double upperValue = Yvals[Yvals.Length - 1];
+            update[Yvals.Length - 1] = upperValue;
+
+            for (int i = Yvals.Length - 2; i >= 0; i--)
+            {
+
+                if (Yvals[i] >= upperValue)
+                {
+                    update[i] = upperValue - epsilon;
+                    upperValue -= epsilon;
+                }
+                else
+                {
+                    update[i] = Yvals[i];
+                    upperValue = Yvals[i];
+                }
+            }
+            Yvals = update;
+        }
+        public void ForceStrictMonotonicityBottomUp(double max = double.MaxValue, double min = double.MinValue)
         {
             double epsilon = 0.005;
             double previousYval = min;
@@ -398,7 +422,6 @@ namespace HEC.FDA.Model.paireddata
             }
             Yvals = update;
         }
-
         public void SortToIncreasingXVals()
         {
             Array.Sort(Xvals,Yvals);
--- a/HEC.FDA.Model/paireddata/PairedData.cs
+++ b/HEC.FDA.Model/paireddata/PairedData.cs
@@ -9,6 +9,7 @@ namespace HEC.FDA.Model.paireddata
     public class PairedData : IPairedData
     {
         #region Fields 
+        private const double EPSILON = double.Epsilon;
         #endregion
 
         #region Properties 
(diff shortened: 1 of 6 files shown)


rnugent3:
Our test is failing but I can't debug the test because ACE-IT said no

rnugent3:
Made this a draft until I have access to the model test project on my machine, ACE-IT permitting

rnugent3:
many appear to be failing, I will hold off until ACE-IT lets me run the tests on my machine

rnugent3:
two problems I am tracking, first, the test on standard deviation is not effective for flow dists, second, despite the ineffective test, I can see that our flow quantile in 2.0 does not match 1.4.3 that well. I am not sure how well it should match. probably have no more room for math today.........

rnugent3:
The integration test in `GraphicalUncertainPairedDataTests` titled `SamplePairedDataShould` is not great. We don't have a great benchmark - we're using 1.4.3 which is a black box, and we can only see inside that box with a narrow flashlight. We may not be tracking logic for the implementation of graphical uncertainty in 1.4.3. I've set 1.4.3 as the benchmark. You can see in the following images that 2.0 is now matching 1.4.3 much better in many cases. In other cases, 2.0 already matched 1.4.3 nicely. 

There is the tail end of the curve that tends not to match between 2.0 and 1.4.3, even after the work in this PR. At this time, I am not able to explain the difference. 

## Algiers (stage) - the origin of this work 
image
image
image

## Glendive (stage) 
image
image
image

## Tafuna 
image
image
image

## London Orleans 
image
image
image

rnugent3:
In those integration tests, we're checking for the error between the solid green line and dotted green line. There is some error.

HEC.FDA.ModelTest/unittests/extensions/GraphicalTests.cs
```diff
@@ -54,60 +55,31 @@ public void ReturnSameNumberOfProbabilitesAsDistributions(double[] probs, double
             Assert.Equal(dists.Length, prob.Length);
         }
 
-        /// <summary>
-        /// Test data: https://docs.google.com/spreadsheets/d/1GhRe3ECAFIKgRqEE8Xo6f_0lHYnHqUW0/edit?usp=sharing&ouid=105470256128470573157&rtpof=true&sd=true
-        /// </summary>
         [Theory]
-        [InlineData(new double[] { 0.999, 0.5, 0.2, 0.1, 0.02, 0.01, 0.005, 0.001, 0.0001 }, new double[] { 80, 11320, 18520, 23810, 35010, 39350, 42850, 47300, 52739.48924 }, 50, true, new double[] { 936.69, 1951.72, 3213.97, 6066.15, 7900.75, 7900.75} )]
-        [InlineData(new double[] { 0.999, 0.5, 0.2, 0.1, 0.02, 0.01, 0.005, 0.001, 0.0001 }, new double[] { 80, 11320, 18520, 23810, 35010, 39350, 42850, 47300, 52739.48924 }, 50, false, new double[] {.2244,  .1164, .1293, .1924, .2030, .2030})]
-        public void ReturnsCorrectStandardDeviations(double[] exceedanceProbabilities, double[] flowOrStageValues, int equivalentRecordLength, bool usingStagesNotFlows, double[] expectedSD)
```
rnugent3, line 63: the logic tested in this method is tested in the graphical uncertain paired data sampled paired data test

HEC.FDA.Model/paireddata/PairedData.cs
```diff
@@ -366,9 +367,30 @@ public void ForceMonotonicity(double max = double.MaxValue, double min = double.
             }
             Yvals = update;
         }
-        public void ForceStrictMonotonicity(double max = double.MaxValue, double min = double.MinValue)
+        public void ForceStrictMonotonicityTopDown(double max = double.MaxValue, double min = double.MinValue)
+        {
+            double[] update = new double[Yvals.Length];
```
Brennan1994, line 372: This is low enough in the stack, (Litteral Billions of paired data are created for a given compute). I want to make sure we're being real memory stingy. I think we could do this in place rather than allocating a new array and then copying it back over.  

If that doesn't make immediate sense, lets talk about it in the office.

Brennan1994 (changes requested):
Lets do some workshopping for performance. see comments

Brennan1994 (approved):
The changes included with this PR are reasonable. i've verified for Graphical Frequency curves, the curve is extrapolated  to 0.9999, and 0.0001. then interpolated to include ordinates at a predetermined intervals before calculating uncertainty distributions around those ordinates. 

in the sampling step, samples above .5 now enforce a top down montonicity instead of a bottom up. 

These two conditions: the forcing of monotonicity and order of operations are the focus of this PR, and the code appears by my review and testing to work as intended. 

All additions and modifications of mine address readability, performance, and convention.

PR #1227, opened 2025-04-01
Unit Test Cleanup

Brennan1994:


HEC.FDA.ModelTest/integrationtests/DefaultDataComputeOutcomes.cs
```diff
@@ -268,7 +268,7 @@ public void AnalyticalWithRegUnreg_ScenarioResults(double expectedResidentialMea
             //Act
             double actualResidentialMeanEAD = scenarioResults.MeanExpectedAnnualConsequences(impactAreaID1, residentialDamageCategory);
             double actualCommercialMeanEAD = scenarioResults.MeanExpectedAnnualConsequences(impactAreaID1, commercialDamageCategory);
-            double tolerance = 0.19;
+            double tolerance = 0.2;
```
rnugent3, line 271: Should we reconsider whether this test is helpful?

HEC.FDA.ModelTest/unittests/StageDamageShould.cs
```diff
@@ -256,32 +256,13 @@ private Inventory CreateInventory()
             return inventory;
         }
 
-
-        [Fact]
```
rnugent3, line 260: do we have a similar test somewhere?

HEC.FDA.ModelTest/unittests/StructureDataValidatorShould.cs
```diff
@@ -8,9 +10,20 @@
 using Xunit;
 
 namespace HEC.FDA.ModelTest.unittests;
+[Trait("RunsOn", "Remote")]
 public class StructureDataValidatorShould
 {
-    static PointShapefile pointShapefile = new PointShapefile(Resources.StringResourcePaths.pathToNSIShapefile);
+    private static PointFeatureCollection pointShapefile = GetFeatures(Resources.StringResourcePaths.pathToNSIShapefile);
```
rnugent3, line 16: nice

rnugent3 (commented):
Looks good. A couple questions to think about.

HEC.FDA.ModelTest/unittests/StageDamageShould.cs
Brennan1994: not necessarily. Truth is this test hadn't been run in a long time, and the case it was testing throws an exception way before it looks like it did in this test. A lot of processing is kicked off right in the constructor of IASD. We might reconsider that, because constructors can't fail as gracefully as we can on other methods. 

For now, this test just isn't useful, and should go.

HEC.FDA.ModelTest/integrationtests/DefaultDataComputeOutcomes.cs
Brennan1994: Likely. Where did the expected values come from?

Brennan1994:
Documented remaining comments in separate issues. Merging because the overall benefit of having these tests updated is necessary to keep the next release on schedule.

PR #1228, opened 2025-04-14
Fix bug in CSVRs where central tendencies and std.dev were swapped

Brennan1994:
Fixes bug where triangular inputs were not being properly translated from 1.4.x formats and mean and standard deviation were  being swapped for normal dists.

### DIFF
--- a/HEC.FDA.ViewModel/Utilities/ImportFromFDA1Helper.cs
+++ b/HEC.FDA.ViewModel/Utilities/ImportFromFDA1Helper.cs
@@ -715,6 +715,15 @@ namespace HEC.FDA.ViewModel.Utilities
                     double newLogStDev = logMean * logStDev / 100;
                     retval = new LogNormal(logMean, newLogStDev);
                     break;
+                case IDistributionEnum.Triangular:
+                    Triangular triDist = dist as Triangular;
+                    double mostLikely = triDist.MostLikely;
+                    double min = triDist.Min;
+                    double max = triDist.Max;
+                    double newMin = mostLikely * min / 100;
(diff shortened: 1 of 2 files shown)


Brennan1994:
Needs another update. There's an inconsistency between Normal and Triangular we need to work out.

HEC.FDA.ViewModel/Utilities/ImportFromFDA1Helper.cs
```diff
@@ -715,6 +715,15 @@ private static ContinuousDistribution ConvertRatioValueToFDA2(ContinuousDistribu
                     double newLogStDev = logMean * logStDev / 100;
                     retval = new LogNormal(logMean, newLogStDev);
                     break;
+                case IDistributionEnum.Triangular:
```
rnugent3, line 718: does it make sense to **rename these variables** to remind us that the min that we import is actually min as a percent of most likely, and that max is max as a percent of most likely?

rnugent3 (commented):
Formatting comment only. Testing import of occupancy type data is a need, I think that could be covered in a manual test plan.

HEC.FDA.ViewModel/Utilities/ImportFromFDA1Helper.cs
Brennan1994: Comments here could and should be better as a whole. i will make some edits

Brennan1994:
Added tickets to investigate documenting/testing/refactoring this class in particular. 

Issue #1232 
Also identified potential bug and logged in #1233

rnugent3 (approved):

PR #1229, opened 2025-04-18
Enhancment graphical visualization

Brennan1994:


rtle76 (approved):
I understand a lot of it, but a lot of it I don't

rtle76:
Appears that the wrong line is solid black:
image

rtle76 (commented):
The lower confidence limit line is solid black, and the other two are dashed blue. Recommend changing it so that the middle line is black and the lower and upper confidence limits are dashed blue.

image

Both tables are editable which could create confusion about which is the input table. Consider indicating explicitly which table is the input table.

There is something going on with exceedance probabilities that are generated in the table on the right - see image below. There are three exceedance probs at .001 with different flow values.

image

rtle76:
I've reviewed latest changes and confirm comments have been addressed

PR #1237, opened 2025-04-30
Bug ias plots

Brennan1994:


rnugent3 (changes requested):
### Changes Requested 
1. The discount rate shows up in several locations, we should update the format for all locations. For example:
image
2. I am not able to launch the alternative comparison report results using the attached study data with the portable build from this PR's action. The software crashes. I thought it might be to do with the two scenarios that don't have results or results are outdated (not sure how we're handling that situation), but seems not to be the case - I computed the scenario and still the software wants to crash. 
image
[WS6_Part2_Solution.zip](https://github.com/user-attachments/files/20018067/WS6_Part2_Solution.zip)
3. The stage-frequency plot in the preview compute remains inverted:
image

### Updated Functionality Working As Expected
1. The study description is retained as expected:
image
2. The preview compute plots with flow-frequency look great:
image

Brennan1994:
The preceding collection of commits fixes the issues identified in review
- Discount rate now shows to 4 decimal places in all controls
- IAS Plot now displays in appropriate orientation and unit for both frequency plots.
- Alternative Comparison Report plot no longer throws exception by default. Robust to failing alternatives.

rnugent3 (changes requested):
I hit an unhandled exception when I attempt to preview compute with just a stage-frequency

image

Brennan1994:
Having trouble reproducing with my datasets. Lets look together tomorrow and figure it out

rnugent3 (approved):
Great work, @Brennan1994 

Run it!!

PR #1238, opened 2025-04-30
Bug/remove view thresholds

Brennan1994:


rnugent3 (changes requested):
The code changes look good. I see great comments, unnecessary usings removed, improved consistency with casing. Thank you @Brennan1994 

I am tracking that work on the study properties window is the only difference between this branch and the other active branch, and I have reviewed/tested the other active branch. 

I have one request on study properties for your consideration. The save button is now available when the window is small on tiny monitors, but the general problem remains: a tiny window prevents access to some of the study properties window. I think some scrolling capability would be helpful. 

For example, take a look at the screen shot below. When the window is small, the save button remains accessible, but convergence criteria is not accessible and can only be accessed with a big window. Ability to scroll would fix the issue. 
image

Brennan1994:
Excellent observation. Can do.

Brennan1994:
You're comment is addressed.

rnugent3 (approved):
Final update looks great. THanks, @Brennan1994 

image

PR #1253, opened 2025-05-29
Lifeloss Investigation

jackschonherr:
Added a new visual scratch project for exploring new UI stuff. Created life loss importer user control which puts data into dynamic histograms


HEC.FDA.Statistics/Histograms/DynamicHistogram.cs
```diff
@@ -131,9 +131,22 @@ private DynamicHistogram(double min, double max, double binWidth, Int64 sampleSi
             ConvergenceCriteria = convergenceCriteria;
             SampleSize = sampleSize;
         }
+        public DynamicHistogram(double min, double binWidth, Int64[] binCounts, double sampleMean, double sampleVariance, double sampleMin, double sampleMax, ConvergenceCriteria convergenceCriteria)
```
Brennan1994, line 134: Use "long" rather than Int64. There's nothing wrong with Int64, and long is just an alias for it, but it's more conventional.

HEC.FDA.Model/metrics/StudyAreaConsequencesBinned.cs
```diff
@@ -60,6 +60,7 @@ internal void AddNewConsequenceResultObject(string damageCategory, string assetC
             AggregatedConsequencesBinned newDamageResult = new(damageCategory, assetCategory, convergenceCriteria, impactAreaID);
             ConsequenceResultList.Add(newDamageResult);
         }
+        // REALLY LOOKS LIKE WE SHOULD BE ADDING THE RESULT TO CONSEQUENCE RESULT LIST HERE - TODO - BBB
```
Brennan1994, line 63: Make an issue on the repo to look at this line of code, and remove this comment. We'll address this once the outstanding 2 pull requests are in.

VisualScratchSpace/ViewModel/CheckBoxImporterVM.cs
Brennan1994, line 1: Very cool. obviously some hard coded paths in here, but that's what a scratch space is for.

VisualScratchSpace/ViewModel/CheckBoxImporterVM.cs
```diff
@@ -0,0 +1,214 @@
...
+        if (_lifeLossFunctions.IsEmpty()) return;
+        _plotIndex = (_plotIndex + 1) % _lifeLossFunctions.Count;
+        ChangePlot(_plotIndex);
+    }
+
+    [RelayCommand]
+    public void SavePlots()
+    {
+        string dbpath = @"C:\FDA_Test_Data\WKS20230525\WKS20230525\save-test.db";
+        LifeLossPlotSaver ps = new(dbpath);
```
Brennan1994, line 84: Nomenclature thing here. We don't actually save plots, because plots are pictures of data. We actually save the backing data, which I might call a relationship, or a function. Some preference for function. Maybe LifeLossFunctionSaver?

VisualScratchSpace/Model/Simulation.cs
```diff
@@ -0,0 +1,24 @@
+﻿using RasMapperLib;
+namespace VisualScratchSpace.Model
+{
+    /// <summary>
+    /// Simple data structure representing the fields which comprise a LifeSim simulation
+    /// </summary>
+    public class Simulation
```
Brennan1994, line 7: To avoid nomenclature confusion. Maybe call this LifeSimSimulation, since we also have a concept of a simulation.

VisualScratchSpace/ViewModel/CheckBoxImporterVM.cs
Brennan1994, line 84: Gut check says there ought to be a using statement around here where we're hitting disk,. Is there an IDisposable here that might not dispose?

VisualScratchSpace/Model/LifeLossFunctionGenerator.cs
```diff
@@ -0,0 +1,131 @@
...
+/// <summary>
+/// Creates a list of life loss functions for each summary zone in simulation for a given set of alternatives and hazard times
+/// </summary>
+public class LifeLossFunctionGenerator
+{
+    private readonly LifeLossDB _db;
+    private readonly LifeLossPlotSaver _saver = new(@"C:\FDA_Test_Data\WKS20230525\WKS20230525\save-test.db");
+    private readonly Dictionary<string, string> _hydraulicsFolderByAlternative;
+    private readonly string _summarySetName;
+    private Dictionary<string, PointM> _indexPointBySummaryZone; // not readonly because we reassign its pointer in CreateLifeLossFunctions, too costly to do in constructor?
```
Brennan1994, line 20: Tell me more what you're thinking from the comment here.

VisualScratchSpace/Model/LifeLossFunctionGenerator.cs
```diff
@@ -0,0 +1,131 @@
...
+        foreach (LifeLossFunction llf  in existingFunctions)
+        {
+            for (int i = 0; i < llf.AlternativeNames.Length; i++)
+            {
+                seenStages[(llf.SimulationName, llf.SummaryZone, llf.AlternativeNames[i])] = llf.Data.Xvals[i];
+                seenHistograms[(llf.SimulationName, llf.SummaryZone, llf.AlternativeNames[i], llf.HazardTime)] = (DynamicHistogram)llf.Data.Yvals[i];
+            }
+        }
+
+        bool newEntries = false; // if this flag remains false, we can just return the existing functions because we only read functions already in the DB
```
Brennan1994, line 92: There's some workflow details here I'm not quite clear on. We should hash out what this workflow ought to look like. and document that somewhere. I'm not totally sure we need to be in a position to update functions already in the database, rather than just delete the old and import the new.

VisualScratchSpace/Model/LifeLossDB.cs
Brennan1994, line 1: LifeSimDatabase may be a better name here, since it is specific to how LifeSim saves it's data.

Brennan1994 (changes requested):

HEC.FDA.Statistics/Histograms/DynamicHistogram.cs
jackschonherr: done. I used it because it is used all over the DynamicHistogram class, do we want to change that?

HEC.FDA.Model/metrics/StudyAreaConsequencesBinned.cs
jackschonherr: done

VisualScratchSpace/Model/Simulation.cs
jackschonherr: renamed.

VisualScratchSpace/Model/LifeLossFunctionGenerator.cs
jackschonherr: I left this here because I have to query the summary zones to ask which of the index points is theirs. I was unsure if that is too costly of an operation to do in the constructor (want those to be lightweight ideally). It is not a readonly property because I reassign its pointer when I query the polygons.

This could all become obsolete once integrating into FDA if there is already an association between summary zones and index points.

VisualScratchSpace/Model/LifeLossFunctionGenerator.cs
jackschonherr: This was created as a layer of security, but if the actual workflow does not present any danger, it can be workshopped.

VisualScratchSpace/Model/LifeLossDB.cs
jackschonherr: renamed.

VisualScratchSpace/Model/LifeLossFunctionGenerator.cs
Brennan1994: Understood. I think this will change enough on actual integration I don't want to get too nitty gritty here. Since this is in scratch space this is just preliminary, and I think this is great.

VisualScratchSpace/Model/LifeLossFunctionGenerator.cs
Brennan1994: Understood. I think this will change enough on actual integration I don't want to get too nitty gritty here. Since this is in scratch space this is just preliminary, and I think this is great.

Brennan1994 (approved):
Approved. Code changes are largely scratch space. Concerns all addressed in person, or recorded in the review here.

PR #1259, opened 2025-08-29
fix #1252

### LINKED ISSUE #1252: Total Risk Configuration Issue
This GitHub issue relates to the user-reported problem with computing Total Risk on Discourse: https://discourse.hecdev.net/t/troubleshooting-total-risk-and-failure-eads-being-exactly-the-same/4143

It appears that the Scenarios configuration identified by the user is not consistently reaching the model. The study data is on the share drive in the HEC-FDA study dropoff folder. Garden City Version 8. 

You can download the data, observe the Scenarios summary results, call those results Set A. note that LS Total Risk and LS Breach have the exact same EAD for impact area 7. You can simply hit recompute, and notice that the results don't change, the results are still Set A. Then, change impact area 7 levee to something else, recompute, notice that results differ, call those results Set B. Then, change impact area 7 levee back to the original configuration, and recompute. We should get summary results Set A, but we get something entirely different, Set C. Set C turns out to be the correct looking results. 

A related configuration issue: we need to make sure that we're not calculating total risk for impact areas that do not have assigned levees within a total risk scenario. So in this case study, we need to be sure that EAD in the Total Risk scenario for impact area 10 is calculated correctly, and I think that means using the failure hydraulics - maybe we should think that one through/talk with the PCX. How to handle the non-leveed impact areas in a Total Risk Scenario.

Brennan1994:


Brennan1994:
The principal issue here was that whether we're computing total risk, and if so, with which damage curves, is a setting stored at the simulation level, not the IAS level, but is information needed when we call IAS to compute. 

In the future, I think that setting ought to be moved down. Having it be scenario level makes for some less intuitive design. This solution sends those setting into the IAS on load, and adds a validation check and hard fail if this somehow ever pops up again.

PR #1266, opened 2025-09-24
Fixes #1265 with validation added to AEP with Given Assurance Method

### LINKED ISSUE #1265: AEP with Assurance Reported as 1.0001
For studies where the threshold stage is super low relative to the frequency function and all AEPs at the threshold stage are about 0.9999, the assurance histogram can return the 90th percentile of AEPs (the AEP that has 90% assurance) as 1.001, which is a nonsensical result. 

Image

In this example, the threshold stage is lower than the 2.5th percentile of the .9999 AEP stages. We are really talking about extreme edge cases. 

Image

rnugent3:
@Brennan1994 this fix is ready for your review. 

This fix does not attempt to improve histogram binning, which is a complex effort left to a later date, probably part of our R&D work unit when we look at empirical hazard uncertainty. 

For now, I added validation where this specific issue will come up. 

I look forward to your thoughts.


rnugent3:
This fix achieves the intended result: 
image

rnugent3:
Failing tests have to do with GDAL. `error NU1101: Unable to find package Geospatial.GDALAssist. `

PR #1271, opened 2025-10-27
Lifeloss integration

jackschonherr:
Life Loss integration prototype

### DIFF
--- a/HEC.FDA.View/Utilities/CustomHeader.xaml
+++ b/HEC.FDA.View/Utilities/CustomHeader.xaml
@@ -104,6 +104,7 @@
               Source="{Binding Path=ImageSource, Converter= {StaticResource NullImageConverter}}"
               Visibility="{Binding Path=ImageSource, Converter={StaticResource NullToCollapsed}}"
               Width="20"
+              Height="20"
               Margin="0,0,5,0" />
             <TextBlock
               Grid.Column="1"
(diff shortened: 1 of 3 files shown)


HEC.FDA.Model/LifeLoss/LifeSimDatabase.cs
```diff
@@ -0,0 +1,144 @@
...
+                    // Hazard Times are in the 2nd through 12th selections                                                           
+                    for (int i = 2; i <= 12; i++)
+                    {
+                        if (reader.GetBoolean(i))
+                            simulation.HazardTimes.Add($"{(i - 1) * 2}"); // convert time to integer string, starting from 2 and incrementing by 2 until 22
+                    }
+                    simulations.Add(simulation);
+                }
+            }
+            catch (Exception ex)
```
Brennan1994: What should we do with this exception? either we should just throw, or handle it somehow. What do you think?

HEC.FDA.View/Resources/lifesim.png
Brennan1994: Funny how nice it looks in the UI. We may ask woody for a nice SVG before we ship.

HEC.FDA.ViewModel/LifeLoss/IndexPointsLifeLossVM.cs
Brennan1994: Take a look at the relationship between GraphicalVM and FrequencyEditorVM. Burying the graphical VM one deeper meant I could update the syntax to use the MVVM toolkit tooling, without violating the existing conventions of BaseViewModel. 

Not mandatory, But it's something to consider here.

Brennan1994 (changes requested):
Recognizing this as two major work in progress features. the Occtype import/export, and LifeLoss.

HEC.FDA.ViewModel/LifeLoss/IndexPointsLifeLossVM.cs
jackschonherr: I did this because this VM needs to access the study cache, which is part of the base view model. It relies on other elements in FDA (hydraulics and index points), so it needs to access the study cache. For example, if the user imports new hydraulics, this VM needs to know about that so it can update its list of hydraulics to select from.

Brennan1994 (approved):

PR #1277, opened 2025-11-06
UI enhancements

### LINKED ISSUE #1272: Tree Elements should highlight (focus) when moused over in FDA
elements in the tree should lightly highlight as you mouse over them to indicate which you're selecting

jackschonherr:
Adds blue focus highlighting to study tree elements. Fixes #1272

### DIFF
--- a/HEC.FDA.View/Study/StudyView.xaml.cs
+++ b/HEC.FDA.View/Study/StudyView.xaml.cs
@@ -117,6 +117,11 @@ namespace HEC.FDA.View.Study
                 return;
             }
 
+            if (e.ChangedButton != MouseButton.Left)
+            {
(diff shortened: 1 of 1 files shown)


Brennan1994:
image

Couple comments. 
- The select highlight is different than the hover highlight. One includes the empty image gap, the other doesn't. 
- The hover highlight has rounded corners, and select doesn't which makes them line up a little weird. 

Neither of these comments impact performance or function, but would be nice polish on the feature add. 
Thanks Jack!

Brennan1994 (changes requested):
Let me know what you think about the comments above. We can either address those now, or backlog them and move forward. Just want to document it here.

PR #1279, opened 2025-11-10
Enhancement model object duplication

### LINKED ISSUE #1024: Copy Study Data Element
It would be really helpful for the user to be able to copy a given element in the study tree (such as a scenario or set of stage damage functions) and duplicate the object to be tailored to their needs. 

This will help cut down set-up time so the user doesn't have to click quite as many frequency functions. For example, I could duplicate a set of stage-damage functions for the without-project condition, duplicate that set, choose the with-project structure inventory, and re-compute, without having to select the same 100 frequency functions I did for the without-project stage-damage functions.

jackschonherr:
Adds the ability to duplicate up to ten model objects at once, complete with a Windows-like naming scheme. Fixes #1024

### DIFF
--- a/HEC.FDA.ViewModel/Utilities/DuplicateVM.cs
+++ b/HEC.FDA.ViewModel/Utilities/DuplicateVM.cs
@@ -16,6 +16,7 @@ public partial class DuplicateVM : BaseEditorVM
         set { _selectedNumDuplicates = value; NotifyPropertyChanged(); }
     }
 
+    // source to display the selectable numbers in the UI
     public int[] NumberOfDuplicateOptions
     {
         get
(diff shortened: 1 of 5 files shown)


HEC.FDA.View/ViewWindow.xaml.cs
```diff
@@ -55,19 +56,32 @@ private void WindowSpawner(WindowVM newvm, bool asDialogue)
                 newwindow.MainGrid.RowDefinitions[0].Height = new GridLength(0);
             }
 
+            SetWindowStartPosition(newwindow);
```
Brennan1994: Clever.

HEC.FDA.ViewModel/Utilities/DuplicateVM.cs
```diff
@@ -0,0 +1,85 @@
...
+        do
+        {
+            candidate = $"{baseName} ({i})";
+            i++;
+        } while (SiblingNameExists(candidate, siblings));
+        lastIndex = i;
+        return candidate;
+    }
+
+    private static bool SiblingNameExists(string name, List<ChildElement> siblings)
```
Brennan1994: This is small enough that it doesn't matter much, but rather than looping through the list, you could convert to a hash set, which would be faster.

jackschonherr: the search would be O(1), but converting this list to a hash set would require O(n) time and an additional allocation. I don't think there's a need to do that as this check only runs once

Brennan1994 (approved):

PR #1292, opened 2025-12-10
Debug tools

Brennan1994:
This PR adds some nice debugging tools, including ScottPlots and an ASCII histogram visualizer. 

It also Rips out the less simple method into a static method in a separate class. Doing this for ease of review. I also removed some parameters that used to be configurable from it, because I feel philosophically they shouldn't be. That method is to be as rigidly the same as 1.4 as possible. 

Critically, computation responsibility moved form GraphicalDistribution -> GraphicalFrequencyCalculators. Those two classes should be the beef of this review.

### DIFF
--- a/HEC.FDA.Model/paireddata/InterpolateQuantiles.cs
+++ b/HEC.FDA.Model/paireddata/InterpolateQuantiles.cs
@@ -27,9 +27,9 @@ namespace HEC.FDA.Model.paireddata
             {
                 neededZs[i] = standardNormalDistribution.InverseCDF(1-exceedanceProbabilitiesForWhichQuantilesAreRequired[i]);
             }
-            IPairedData nonexceedance_zScore = new PairedData(exceedanceProbabilitiesForWhichQuantilesAreRequired, neededZs);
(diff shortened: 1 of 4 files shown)


HEC.FDA.Model/paireddata/UncertainPairedData.cs
```diff
@@ -114,14 +114,29 @@ private static bool IsDistributionArrayValid(IDistribution[] arrayOfData, double
...
+        /// returns a paired data sampled at the given probability without enforcing any monotonicity
+        /// </summary>
+        public PairedData SamplePairedDataRaw(double probability)
+        {
+            double[] y = new double[Yvals.Length];
+            for (int i = 0; i < Xvals.Length; i++)
+            {
+                y[i] = Yvals[i].InverseCDF(probability);
+            }
+            PairedData pairedData = new(Xvals, y, CurveMetaData);//mutability leakage on xvals
```
jackschonherr: what does mutability leakage mean here? Xvals is readonly from what I can tell

jackschonherr (commented):
Not much on this review, code looks good

jackschonherr (approved):
LGTM

PR #1293, opened 2025-12-11
Add Fn Plot Control LL

Brennan1994:
* **Initial implementation of the FN Plot that is defined in TRG-1**
image

### DIFF
--- a/HEC.FDA.ViewModel/LifeLoss/LifeLossFnChartVM.cs
+++ b/HEC.FDA.ViewModel/LifeLoss/LifeLossFnChartVM.cs
@@ -121,7 +121,7 @@ public class LifeLossFnChartVM : BaseViewModel
         var xAxis = new LogarithmicAxis
         {
             Position = AxisPosition.Bottom,
-            Title = "Average Life Loss",
+            Title = "Average Annual Life Loss",
             Minimum = 0.1,
             Maximum = 10000,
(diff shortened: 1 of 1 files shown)


HEC.FDA.ViewModel/LifeLoss/LifeLossFnChartVM.cs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    /// <summary>
+    /// Updates the chart with new data, clearing existing data series but keeping zone lines.
+    /// </summary>
+    /// <param name="data">Pre-transformed UncertainPairedData where X = Average Life Loss, Y = AEP distribution</param>
+    public void UpdateData(UncertainPairedData data)
+    {
+        // Remove data series but keep zone boundary lines (first 2 series)
+        while (PlotModel.Series.Count > 2)
```
jackschonherr: Are we guaranteeing that the boundary zone lines will always be the first two series in this collection?

jackschonherr (commented):
Code looks good. Just a couple questions

HEC.FDA.ViewModel/LifeLoss/LifeLossFnChartVM.cs
Brennan1994: Modified. we now add a tag to those series, so we can more intelligently ask for them,

PR #1320, opened 2026-08-04
refinements

Brennan1994:


HEC.FDA.ViewModel/ImpactAreaScenario/Editor/IASEditorVM.cs
```diff
@@ -258,13 +312,15 @@ private void FillForm(IASElement elem)
                 {
                     SpecificIASEditorVM specificIASEditorVM = new SpecificIASEditorVM(foundElement, row, GetSelectedStageDamage);
                     specificIASEditorVM.RequestNavigation += Navigate;
+                    specificIASEditorVM.PropertyChanged += SpecificIAS_PropertyChanged;
```
jackschonherr, line 315: I don't think existing events are unsubscribing in general in these classes, so I think this is ok but just wanted to note

jackschonherr (approved):
LGTM, only one comment about not unsubscribing from events, but it follows precedent in these files

HEC.FDA.ViewModel/ImpactAreaScenario/Editor/IASEditorVM.cs
Brennan1994: I think that's right. worth knowing ourselves in case memory issues pop up some day, but this is status quo for now.

